PR #50, opened 2017-01-11
Add analytics events

lilaconlee:
This adds Drop-in specific analytics events in addition to events sent from the client, hosted fields, and paypal components. 

Events include: 

| Event name  | When does this happen |
|----|----|
| web.dropin.started.tokenization-key  | When Drop-in is started, with tokenization key |
| web.dropin.started.client-token  | When Drop-in is started, with client token |
| web.dropin.appeared  | When Drop-in appears after inital loading screen with async dependencies |
| web.dropin.selected.card   | When "Card" is selected from the payment options list |
| web.dropin.selected.paypal   | When "PayPal" is selected from the payment options list |
| web.dropin.error.configuration-error   | When Drop-in is instantiated with an invalid configuration |
| web.dropin.request-payment-method.error  | When requestPaymentMethod is called and no payment method is avaliable |
| web.dropin.request-payment-method.card   | When requestPaymentMethod is called with a CreditCard |
| web.dropin.request-payment-method.paypal   | When requestPaymentMethod is called with a PayPalAccount |

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -38,6 +38,6 @@ module.exports = {
   },
   ANALYTICS_REQUEST_TIMEOUT_MS: 2000,
   ANALYTICS_PREFIX: 'web.dropin.',
-  INTEGRATION: 'dropin',
+  INTEGRATION: 'dropin2',
   STYLESHEET_ID: 'braintree-dropin-stylesheet'
 };


casim:
hi guys,
is there any way to access these analytics? we would like to know more about what are clients are doing on the drop-in component.

PR #62, opened 2017-02-09
Disable Card entry when no supported card types

quinnjn:
Currently, we always show the Cards payment method. If the merchant has no supported card types, the customer would not be able to use any card. 

Fixed by checking if the merchant can support any card types before showing supporting the card payment option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 unreleased
 ----------
 - Use version 3.8.0 of braintree-web
+- Doesn't show Card payment option for merchants without cards enabled
 
 1.0.0-beta.4
 ------------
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ CHANGELOG
 unreleased
 ----------
 - Use version 3.8.0 of braintree-web
+- Fix bug where Card entry would be possible when no cards were supported
```
lilaconlee: I might change this to something along the lines of "Add support for merchants without cards enabled" because this was a known (albeit not documented) behavior.

quinnjn:
Merge onto master: cd43cd41356b3b356877349811d7acc81e1c1fc9

PR #65, opened 2017-02-10
Deprecate set placeholder

lilaconlee:
### Summary
This uses the new `setAttribute` method in Hosted Fields instead of `setPlaceholder`.

### Checklist

~Added a changelog entry~
- [x] Ran unit tests


lilaconlee:
merged with f6e98a1

PR #67, opened 2017-02-10
Swap out paypal component for paypal checkout component

crookedneighbor:
### Summary

This PR swaps out the [PayPal](http://braintree.github.io/braintree-web/current/PayPal.html) component for the [PayPal Checkout](http://braintree.github.io/braintree-web/current/PayPalCheckout.html) component. 

Blockers

- [x] PayPal Checkout.js logs _a ton_ to the console. There's some talk about providing a config option to suppress logs, but it is not active yet. We may not want to include it until then.
- [x] The demo app will not currently work, because the tokenization key is associated with a sandbox account that does not have a linked PayPal account

### Checklist

- [x] Ran unit tests


intelliot:
So the `paypal-checkout` component simply doesn't support the mock sandbox flow (when the sandbox account doesn't have a linked PayPal account) at all?

crookedneighbor:
Correct. Braintree created the mock sandbox flow, so in the old PayPal component, the Braintree GW would provide the url for the PayPal sandbox if it was a linked account and a mock url hosted by Braintree if it was not a linked account. Since the old component was the thing opening the window, it could control this. Now that checkout.js opens the window, we cannot have it open the Braintree mock sandbox.

src/views/payment-sheet-views/paypal-view.js
```diff
@@ -14,24 +16,56 @@ PayPalView.prototype = Object.create(BaseView.prototype);
 PayPalView.prototype.constructor = PayPalView;
 PayPalView.ID = PayPalView.prototype.ID = paymentOptionIDs.paypal;
 
+PayPalView.prototype.setLogLevel = function (paypal) {
+  var level = this.model.merchantConfiguration.paypal.logLevel;
+
+  paypal.setup({
```
crookedneighbor: We discovered that `paypal.setup` cannot be spied on by sinon, so we wrapped it in our own function so we could test the behavior. I don't really like this, but it's the best we could come up with.

crookedneighbor: The reason for this is that it is made so you can't overwrite the setup function.

test/unit/views/payment-sheet-views/paypal-view.js
```diff
@@ -52,64 +62,69 @@ describe('PayPalView', function () {
...
-    it('starts async dependency', function () {
+    it('starts async dependency', function (done) {
       var payPalView;
 
       this.sandbox.stub(DropinModel.prototype, 'asyncDependencyStarting');
 
       payPalView = new PayPalView(this.paypalViewOptions);
 
-      expect(payPalView.model.asyncDependencyStarting).to.be.calledOnce;
+      waitForInitialize(function () {
```
crookedneighbor: Is there a better way to wait for the async stuff to finish?

PR #68, opened 2017-02-10
Transition helper

quinnjn:
### Summary

`onTransitionEnd` will help shim transitions for IE 9. We are expecting transition end events to continue on to the next view, but since IE 9 does not support transitions we need to use the transition helper to give us a callback pattern.

Note that there is no usage of this yet, but will be used on our transition branch. 

### Checklist

- [X] Ran unit tests

### DIFF
--- a/test/unit/lib/unit/transition-helper.js
+++ b/test/unit/lib/unit/transition-helper.js
@@ -1,0 +1,36 @@
+'use strict';
+
+var browserDetection = require('../../../../src/lib/browser-detection');
+var onTransitionEnd = require('../../../../src/lib/transition-helper').onTransitionEnd;
+
+describe('onTransitionEnd', function () {
+  it('immediately calls callback when IE9', function (done) {
+    var element = document.createElement('div');
+
+    this.sandbox.stub(browserDetection, 'isIe9').returns(true);
+
+    onTransitionEnd(element, function () {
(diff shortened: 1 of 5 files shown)


test/unit/lib/unit/transition-helper.js
```diff
@@ -0,0 +1,28 @@
...
+
+    onTransitionEnd(element, function () {
+      done();
+    });
+  });
+
+  it('calls callback after onTransitionEnd end', function (done) {
+    var element = document.createElement('div');
+
+    element.addEventListener = this.sandbox.stub().yields();
```
danakatz: Could we assert that this gets called once, and that it doesn't get called in the previous test?

danakatz:
Also, do we need to use `webkitTransitionEnd` for [older browsers](https://developer.mozilla.org/en-US/docs/Web/Events/transitionend#Browser_compatibility) and, if so, would this be a good place for that logic?

lilaconlee:
@danakatz looks like the browsers requiring `webkitTransitionEnd` are farther back than we support so I think we're good keeping just `transitionend`.

test/unit/lib/unit/transition-helper.js
lilaconlee: 👍 good call

PR #70, opened 2017-02-16
Tokenization animation

quinnjn:
### Summary

We added animations!

## Payment entry to tokenized payment method
tokenized

## Selecting saved payment methods
methodswitch

## Selecting another way to pay
sheetopen

### Checklist

- [X] Added a changelog entry
- [X] Ran unit tests

### DIFF
--- a/src/scss/main.scss
+++ b/src/scss/main.scss
@@ -107,6 +107,7 @@ $curve-springy: cubic-bezier(0.43, 0.48, 0.62, 1.07);
   font-size: 16px;
   color: $text-color;
   margin: 5px 5px 10px 5px;
+  transition: transform 0.2s ease;
 }
 
 .braintree-options {
@@ -167,7 +168,6 @@ $curve-springy: cubic-bezier(0.43, 0.48, 0.62, 1.07);
 
 .braintree-sheet__container {
   margin: 0 auto 13px auto;
-
 }
 
 .braintree-sheet__error {
@@ -234,6 +234,7 @@ $curve-springy: cubic-bezier(0.43, 0.48, 0.62, 1.07);
   display: flex;
   flex-wrap: wrap;
   padding: 12px 15px 0 12px;
+  position: relative;
 
   .braintree-sheet__header-label {
     align-items: center;
@@ -252,7 +253,7 @@ $curve-springy: cubic-bezier(0.43, 0.48, 0.62, 1.07);
   .braintree-sheet__card-icon {
     display: inline-block;
     padding-left: 5px;
-    
+
     &.braintree-hidden {
       display: none;
     }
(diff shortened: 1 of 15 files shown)


test/unit/views/main-view.js
```diff
@@ -605,22 +607,32 @@ describe('MainView', function () {
       });
 
       describe('when a payment sheet is active', function () {
+        var clock;
+
         beforeEach(function () {
-          classlist.add(this.paymentMethodsElement, 'braintree-methods--active');
+          clock = sinon.useFakeTimers();
```
EvanHahn: Minor: could we make this `this.clock`? Saw this in a few other spots, too.

crookedneighbor: Agreed.

src/scss/main.scss
```diff
@@ -499,6 +594,18 @@ $curve-springy: cubic-bezier(0.43, 0.48, 0.62, 1.07);
   display: block;
 }
 
+[data-braintree-id="methods-container"] {
```
lilaconlee: Would slightly prefer having a `braintree-methods-container` class or something along those lines here. We've loosely been trying to keep visual styles associated with class names and the hiding and showing functionality associated with the braintree id data attributes.

We could update the `braintree-options-list` to be `braintree-options-container` to keep this structure for both views, but we can do that as a different PR to keep it in scope.

src/views/main-view.js
```diff
@@ -80,11 +80,14 @@ MainView.prototype._initialize = function () {
 
   this.model.on('changeActivePaymentView', function (id) {
     if (id === PaymentMethodsView.ID) {
-      classlist.add(paymentMethodsViews.element, 'braintree-methods--active');
+      classlist.add(paymentMethodsViews.container, 'braintree-methods--active');
+      classlist.remove(paymentMethodsViews.container, 'braintree-methods-container--inactive');
```
lilaconlee: Does `braintree-methods-container--inactive` get added back in the case that the transition needs to happen again? Didn't see it but I could just be missing it.

quinnjn: It doesn't look like we are using this property anymore. We removed it and rechecked that it didn't effect anything.

If you have an idea on why we should keep this let us know!

src/scss/main.scss
quinnjn: [Related converation](https://github.com/braintree/braintree-web-drop-in/pull/70#discussion_r101663983)

There was only one other property defined under here after cleaning up `braintree-methods-container--inactive`. This property was also defined in `.braintree-loader__indicator`.

So we removed this all together.

lilaconlee:
Updated with a fix for IE and integration tests are passing

quinnjn:
@lilaconlee I'm 👍 on changes in f22f06974726df889e5f0dd85c765c11f4b5b7ba. Hopefully this fixes the integration tests than I think it is ready for 🏅

PR #71, opened 2017-02-21
Prepare package for publishing to npm

EvanHahn:
### Summary

This prepares Drop-in for a publish to `npm`. It fills `dist/npm` with files ready to be `npm publish`ed.

We tested this by doing something like this in another project (and getting all the way to tokenization):

```js
// ...
"dependencies": {
  "braintree-web-drop-in": "/path/to/braintree-web-drop-in/dist/npm"
},
// ...
```

### Checklist

- [x] Added a changelog entry
- [x] Ran unit tests

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -3,6 +3,7 @@
 /* globals __dirname */
 
 var browserify = require('browserify');
+var envify = require('gulp-envify');
 var brfs = require('gulp-brfs');
 var cleanCSS = require('gulp-clean-css');
 var del = require('del');
@@ -114,6 +115,7 @@ gulp.task('build:npm:package.json', function (done) {
   var pkg = Object.assign({}, require('./package.json'));
 
   delete pkg.browserify;
+  pkg.main = 'index.js';
 
   mkdirp.sync(NPM_PATH);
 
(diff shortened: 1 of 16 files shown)


gulpfile.js
```diff
@@ -65,10 +68,6 @@ gulp.task('build:js', ['build:js:unmin', 'build:js:min']);
 
 gulp.task('build:js:unmin', function () {
   return browserify(config.src.js.main, {standalone: 'braintree.dropin'})
-    .transform(stringify, {
```
intelliot: Why did we need `stringify`?

EvanHahn: We have things like `var mainHtml = require('./main.html')`, which we wanted to transform into things like `var mainHtml = '<div class="dropin">...</div>'`.

`stringify` was one way to do it. We found another way that's compatible with npm, which is `brfs`. They accomplish similar things for us.

gulpfile.js
```diff
@@ -107,14 +102,50 @@ gulp.task('build:link-latest', function (done) {
...
+gulp.task('build:npm:statics', function () {
+  return gulp.src([
+    './.gitignore',
+    './CHANGELOG.md',
+    './LICENSE',
+    './README.md'
+  ]).pipe(gulp.dest(NPM_PATH));
+});
+
+gulp.task('build:npm:package.json', function (done) {
```
intelliot: This is clever

PR #73, opened 2017-02-23
Add instructions for getting hosted Drop-in files to README

lilaconlee:
### Summary

Adds some clarification on how to get Drop-in. We'll add npm installation instructions here once it's available. We could also add a bit about how we support browserify/webpack but that might be better suited for the someday dev docs. 

### Checklist

~Added a changelog entry~
~Ran unit tests~

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,14 @@ Because we're still in beta, the API and designs are subject to change. If you h
 - Not in an iframe; feel free to style Drop-in to blend in with your website
 - Open source and open development
 
+## Setup
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,6 +10,14 @@ Because we're still in beta, the API and designs are subject to change. If you h
 - Not in an iframe; feel free to style Drop-in to blend in with your website
 - Open source and open development
 
+## Setup
+
+Drop-in is currently available directly from our servers, which you can save locally or include in your project through a script tag:
+
+```
```
crookedneighbor: nitpick. Use `html` here for fun syntax highlighting.

PR #89, opened 2017-03-07
Add method to update PayPal config options

### LINKED ISSUE #87: Dynamic PP amount?
```
braintree.dropin.create({
            authorization: 'sandbox_XXX ',
            selector: '#braintree-dropin',
            paypal: {
                flow: 'checkout',
                amount : this.getMoney(), // <== THIS
                currency: 'EUR'
            }
        }
```

Is it possible to have the amount property dynamic? At the moment the amount value is bound on construct time, and since we cannot do byref assignment in JS, this is a problem.

I'm looking for a way to do something like `braintree.paypal.amount = XYZ`

At the moment, I am emptying the #braintree-dropin and re-creating it, but that's obviously shit then all CCs data are lost. So, for example, if someone changed his mind about how much they wanna spend/pay, and I reconstruct the dropin and the CC data are lost.

crookedneighbor:
Closes #87

### Summary

Adds a method to dynamically update the PayPal config.

Also clones the merchant config, to prevent tampering with the object after creation.

### Checklist

- [x] Added a changelog entry
- [x] Ran unit tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,8 @@ dropinInstance.setPayPalOption('amount', '20.00');
 dropinInstance.setPayPalOption('intent', null);
 ```
 
+Note: if your customer has already authenticated their PayPal account, using `setPayPalOption` will not make any changes for this session. If the customer chooses to pick another way to pay and authenticates with PayPal again, the changes you make to the PayPal options will be used.
+
 ## Full example
 
 This is a full example of a Drop-in integration that only accepts credit cards.
--- a/README.md
+++ b/README.md
@@ -80,7 +80,11 @@ The structure of the PayPal payment method object returned in the callback of `r
 If you need to update a field (such as the amount) after creation, you can do so with the `setPayPalOption` method.
 
 ```js
+// set amount to 20.00
 dropinInstance.setPayPalOption('amount', '20.00');
+
+// remove intent from options
+dropinInstance.setPayPalOption('intent', null);
 ```
 
 ## Full example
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -77,6 +77,12 @@ You can find more PayPal configuration options in the [Braintree JS client SDK v
 
 The structure of the PayPal payment method object returned in the callback of `requestPaymentMethod` can be found [here](http://braintree.github.io/braintree-web/current/PayPal.html#~tokenizePayload).
 
+If you need to update a field (such as the amount) after creation, you can do so with the `setPayPalOption` method.
+
+```js
+dropinInstance.setPayPalOption('amount', '20.00');
+```
```
lilaconlee: I'd add an example of how to remove an option until we have docs elsewhere.

crookedneighbor: I knew I was forgetting something. Thanks!

intelliot:
What if the customer authenticates their PayPal account and then changes the amount afterward?

The Drop-in UI wouldn't change, but the nonce/authentication will not have been updated to reflect the new amount.

crookedneighbor:
Hmm... that's a good point. 

Do we need to keep track of the state and prevent changing the amount after the user has authenticated?

I'm not actually sure what the behavior is when someone does a transaction.sale call with a different amount.

crookedneighbor:
All issues should be addressed. Please review!

crookedneighbor:
We talked offline and decided that there are some unresolved issues with doing this. Going to close this for now.

PR #99, opened 2017-03-14
Add configuration overrides to test app

crookedneighbor:
### Summary

Adds the ability to pass in a custom configuration in the test app through query params. 

I'm not super worried about xss vectors with this, as it is a test app. If we actually expose it on the internet (IE, Github pages), we should be more worried.

### Checklist

- [ ] ~~Ran unit tests~~
- [x] Ran integration tests

### DIFF
--- a/test/app/index.html
+++ b/test/app/index.html
@@ -113,7 +113,7 @@
       return obj;
     }, {});
 
-    if (config.mergeWithDefault) {
+    if (config.mergeWithDefault !== false) {
       config = Object.assign({}, defaultCreateObject, config);
     }
 
--- a/test/app/index.html
+++ b/test/app/index.html
@@ -65,7 +65,7 @@
 
 <pre id="results"></pre>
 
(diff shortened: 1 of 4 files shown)


test/app/index.html
```diff
@@ -65,31 +65,67 @@
...
+        value = jsonParsedValue;
+
+      } catch (e) {}
+
+      obj[prop] = value;
+
+      return obj;
+    }, {});
+
+    if (config.mergeWithDefault) {
```
crookedneighbor: This should probably be the default behavior. Would probably work better to do:

```js
if (config.mergeWithDefault !== false) {
```

test/app/index.html
```diff
@@ -65,31 +65,67 @@
 
 <pre id="results"></pre>
 
+<script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=Object.assign"></script>
```
crookedneighbor: Had to add a polyfill for `Object.assign`. Do we care which polyfill we use? We could just copy and paste the polyfill from mdn and include as a separate script tag.

PR #109, opened 2017-03-20
Remove duplicate stylesheet by setting the id

intelliot:
### Summary
We were checking for the existence of an element with id
'braintree-dropin-stylesheet' but we failed to actually set the id
anywhere.

The test app was including both a local stylesheet and one from the
assets server. This fixes the issue by not including the assets server
stylesheet when the local one is already present.

### Checklist

- [x] Ran unit tests

### DIFF
--- a/test/unit/dropin.js
+++ b/test/unit/dropin.js
@@ -34,8 +34,14 @@ describe('Dropin', function () {
   });
 
   afterEach(function () {
+    var stylesheet = document.getElementById(constants.STYLESHEET_ID);
+
     if (document.body.querySelector('#foo')) {
       document.body.removeChild(this.container);
+    }
+
+    if (stylesheet) {
+      stylesheet.parentNode.removeChild(stylesheet);
     }
   });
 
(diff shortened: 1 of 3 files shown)


test/unit/dropin.js
```diff
@@ -186,6 +187,16 @@ describe('Dropin', function () {
       }.bind(this));
     });
 
+    it('injects stylesheet with correct id', function (done) {
```
crookedneighbor: Do we have a test already that says we don't inject the stylesheet if a stylesheet with that id already exists?

intelliot: Nope, but we do have [this line of code](https://github.com/braintree/braintree-web-drop-in/blob/9e9e9e61ba48cf5676bfb751ab6356a6ef1ea57b/src/dropin.js#L147)

juongithub:
In a scenario where the CSS is customized, it would be useful to simply check if a CSS property is loaded:

JavaScript:
`function isBraintreeCSSLoaded() {
    var element = document.createElement('div');
    element.className = 'braintree-check';

    if (element.currentStyle) {
        return (element.currentStyle['display'] == 'none');
    } else if (document.defaultView && document.defaultView.getComputedStyle) {
        return (document.defaultView.getComputedStyle(element, null).getPropertyValue('display') == 'none');
    } else {
        return (element.style['display'] == 'none');
    }

    return false;
}`

CSS:
`<style>.braintree-check{display:none}</style>`

test/unit/dropin.js
crookedneighbor: We should have a test around this, I think.

crookedneighbor: I added this test.

PR #111, opened 2017-03-21
Loading animation on "card" element should be stopped on errors

juongithub:
These CSS lines stop the loading animation on "card" element when an error is shown (maybe a code solution would be preferable).


src/scss/main.scss
```diff
@@ -329,6 +329,17 @@ $curve-easeOutBack: cubic-bezier(0.175, 0.885, 0.32, 1.275);
   }
 }
 
+.braintree-sheet--has-error .braintree-sheet--loading {
+  .braintree-sheet__content {
+    opacity: 1.0!important;
```
crookedneighbor: Do these need important declarations?

crookedneighbor:
Can you provide steps for us to reproduce so we can experience the UI problem?

crookedneighbor:
Can you provide steps for us to reproduce so we can experience the UI problem?

juongithub:
1. generate an auth code with verify=true
2. try to use the card 4000111111111115 
3. by calling requestPaymentMethod() you will see an error but the loading animation is still on

src/scss/main.scss
juongithub: actually, "!important" should not be needed, but the dropin script always load the external CSS (#109).

juongithub:
It seems the new commit has solved this problem.

crookedneighbor:
Doesn't seem like this is an issue any longer. Closing.

PR #112, opened 2017-03-22
README - Add section about Custom styling

intelliot:
### Summary
Thanks to #109, developers can use custom styling and prevent Drop-in from loading its own styles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,8 +126,8 @@ This is a full example of a Drop-in integration that only accepts credit cards.
 Customize the look and feel of Drop-in by using your own custom stylesheet.
 
 1. Copy the compiled CSS file to start with our Drop-in style as the foundation.
-2. Make edits to the styles.
-3. Link to your stylesheet. To prevent Drop-in from injecting its own stylesheet, set `id="braintree-dropin-stylesheet"`:
+1. Make edits to the styles.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -121,6 +121,17 @@ This is a full example of a Drop-in integration that only accepts credit cards.
 </html>
 ```
 
+## Custom styling
+
+Customize the look and feel of Drop-in by using your own custom stylesheet.
+
+1. Copy the compiled CSS file to start with our Drop-in style as the foundation.
```
crookedneighbor: Let's use `1.` for every bullet point. Github will render the order correctly.

crookedneighbor: Should we be explicit about where to copy the file?

intelliot: Where would that be?

crookedneighbor: Depends, if you clone down the project and build it, it'll exist at `dist/web/dropin/dev/css/dropin.css`. Otherwise, it'll exist on our cdn, but the version would need to be updated with every release.

lilaconlee:
I think we want to recommend custom styling through editing `data-braintree-id` attributes rather than grabbing our compiled stylesheets, but we can have a more thorough conversation about this offline if we want.

lilaconlee:
Gonna move this to `DEVELOPMENT.md` to indicate how to develop with local CSS and now how we recommend custom styling.

crookedneighbor:
:+1:

PR #121, opened 2017-04-05
Use locale in PayPal Checkout

lilaconlee:
### Summary

We weren't using the locale from the merchant configuration in PayPal Checkout; this fixes that.

### Checklist

~Added a changelog entry~
- [x] Ran unit tests

### DIFF
--- a/src/views/payment-sheet-views/paypal-view.js
+++ b/src/views/payment-sheet-views/paypal-view.js
@@ -35,7 +35,7 @@ PayPalView.prototype._initialize = function () {
 
   btPaypal.create({client: this.client}, function (err, paypalInstance) {
     var paypalCheckoutConfiguration;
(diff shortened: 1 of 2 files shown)


src/views/payment-sheet-views/paypal-view.js
```diff
@@ -61,7 +61,11 @@ PayPalView.prototype._initialize = function () {
       onError: reportError
     };
 
-    paypal.Button.render(paypalCheckoutConfig, '[data-braintree-id="paypal-button"]').then(function () {
+    if (merchantConfiguration.locale) {
```
crookedneighbor: Do we care about the case where they have a different locale in their top level config and their paypal config?

lilaconlee: I'm personally okay with leaving it how it is until someone comes up with a valid use case for differing locales.

PR #126, opened 2017-04-11
Add PayPal Credit

lilaconlee:
### Summary
This adds PayPal Credit to Drop-in with the following API:
```js
+braintree.dropin.create({
+  authorization: 'CLIENT_AUTHORIZATION',
+  selector: '#dropin-container',
+  paypalCredit: {
+    flow: 'checkout', // Required for PayPal Credit
+    amount: 10.00,
+    currency: 'USD'
+  }
+}, callback);
```

The major change is pulling out the good bits of the `PayPalView` into `BasePayPalView` so it can be used for both the PayPal and PayPal Credit flows with minor changes. 

It will appear like so: 

screen shot 2017-04-11 at 11 17 42 am

screen shot 2017-04-11 at 11 19 25 am

We're working on refining the icons for PayPal and PayPal Credit, but that will be done as a separate PR.

### Checklist

- [x] Added a changelog entry
- [x] Ran unit tests


test/unit/views/payment-sheet-views/base-paypal-view.js
```diff
@@ -0,0 +1,482 @@
+'use strict';
+/* eslint-disable no-new */
+/* eslint-disable */
```
crookedneighbor: We don't this was a mistake. Thanks for catching it!

lilaconlee:
@quinnjn I'm okay leaving it off for now, but we're intending to add more configuration options to the demo app that should make it easy to enable.

lilaconlee:
@crookedneighbor mentioned we probably want integration tests, so we'll probably add it by default for now and then make it not default when we actually update the test app.

README.md
```diff
@@ -77,6 +77,26 @@ You can find more PayPal configuration options in the [Braintree JS client SDK v
...
+  selector: '#dropin-container',
+  paypalCredit: {
+    flow: 'checkout', // Required for PayPal Credit
+    amount: 10.00,
+    currency: 'USD'
+  }
+}, callback);
+```
+
+PayPal Credit configuration parameters are the same as those for [PayPal](https://braintree.github.io/braintree-web/current/PayPal.html#tokenize).
```
crookedneighbor: Should we link to the PayPal Checkout component. There is a very slight difference here.

lilaconlee: Yup there and on line 76 probably

PR #132, opened 2017-04-20
Add tab navigation support

crookedneighbor:
### Summary

Adds support for tabbed navigation in drop-in.

*Note:* The PayPal buttons can be tabbed to and accessed by pressing enter, but they will not be highlighted. A change needs to be made in checkout.js for that to work.

### Checklist

- [x] Added a changelog entry
- [x] Ran unit tests

### DIFF
--- a/src/lib/set-up-event-handlers.js
+++ b/src/lib/set-up-event-handlers.js
@@ -1,0 +1,12 @@
+'use strict';
+
+function setUpEventHandlers(element, func) {
+  element.addEventListener('click', func);
+  element.addEventListener('keyup', function (event) {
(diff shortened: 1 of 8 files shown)


crookedneighbor:
tab

src/lib/set-up-event-handlers.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+function setUpEventHandlers(element, func) {
```
skunkworks: I'd like to rename this to something more descriptive, but I'm coming up with a blank. `setUpSelectionEventHandlers`? `addSelectionEventHandler`?

PR #133, opened 2017-04-21
Add fail on duplicate and cvv verification failed errors

crookedneighbor:
### Summary

Updates the card UI to tell the customer that their tokenization attempt was rejected for being a duplicate payment method (if options.failOnDuplicatePaymentMethod is used when generating the client token) or when cvv verification fails.

### Checklist

- [x] Added a changelog entry
- [x] Ran unit tests


src/translations/en.js
```diff
@@ -17,6 +17,8 @@ module.exports = {
   fieldInvalidForPostalCode: 'This postal code is not valid.',
   genericError: 'Something went wrong on our end.',
   hostedFieldsFailedTokenizationError: 'Please check your information and try again.',
+  hostedFieldsTokenizationFailOnDuplicateError: 'This credit card is already saved to a customer.',
```
crookedneighbor: Wasn't sure how to word this. 

When the the failOnDuplicatePaymentMethod option is passed in client token generation, it will fail if the card already exists in the merchant's vault, even if the customer is different.

lilaconlee: I can't think of a better option, but it might be worth checking with the designers?

lilaconlee:
Did we ever come to a conclusion about merchant/UI errors for duplicates? I think I'm still a fan on keeping the generic UI error and could go either way on the merchant error.

intelliot:
I would make sure it's possible for the merchant to show a custom error in their UI in that case. I can easily envision a merchant wanting to show "Credit card already in use" to the user.

PR #135, opened 2017-04-24
Set PayPal button to be translatable

### LINKED ISSUE #134: Missing translation and CSS bug
### General information

* SDK/Library version: 1.0.0 Beta 6
* Environment: Sandbox
* Browser and OS Chrome 57.0.2987.133 (64-bit)) on MacOS 10.12.4

### Issue description

Screenshot:

screen shot 2017-04-24 at 10 47 10 am

Both issues are visible in the screenshot:

- Translation is not correct. The DropIn-UI is set to de_DE (works fine for the other parts)
- CSS is buggy. Element has a max-width of 148px, but the main container is larger by default

crookedneighbor:
closes #134

### Summary

We weren't passing in the locale to the `paypal.Button.render` function, so while the popup was still being translated through the `createPayment` call, the button itself was not.

### Checklist

- [x] Added a changelog entry
- [x] Ran unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ unreleased
 - Add support for PayPal Credit
 - Fix bug where adding a vaulted payment method would duplicate previously added payment methods
 - Add support for tab navigation
+- Fix bug where PayPal button was not being translated
 
 1.0.0-beta.6
 ------------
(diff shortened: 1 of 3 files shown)


crookedneighbor:
The integration test failures are due to an incompatibility with the `locale: 'es_ES'` and `stye: { label: 'credt'}` in checkout.js. I've contacted that team about the bug.

PR #151, opened 2017-05-05
Record Drop-in version in metadata

quinnjn:
### Summary

Record and send the Drop-in version when setting the initial metadata.

### Checklist

- [X] Added a changelog entry
- [X] Ran unit tests


quinnjn:
Closing due to new analytic keys being hard to add.

The `dropinVersion` column doesn't exist in mister_clean and/or redshift and it sounds like a lot of work to get that added which sucks.

We're going to explore setting the Drop-in version in `sdkVersion`

intelliot:
We've decided to pursue this path.

PR #160, opened 2017-05-11
Add timeout for async dependencies in PayPal

lilaconlee:
### Summary

This adds a timeout for loading PayPal dependencies. When PayPal sandbox was down, Drop-in got stuck at the spinner. Now the payment options view loads with PayPal options greyed out with an error message. This message isn't currently translated, but as it was added for sandbox issues I'm fine with adding it as is for now and adding the translations as we get them.

screen shot 2017-05-11 at 11 26 47 am

### Checklist

- [x] Added a changelog entry
- [x] Ran unit tests

### DIFF
--- a/src/views/payment-sheet-views/base-paypal-view.js
+++ b/src/views/payment-sheet-views/base-paypal-view.js
@@ -4,6 +4,8 @@ var BaseView = require('../base-view');
 var assign = require('../../lib/assign').assign;
 var btPaypal = require('braintree-web/paypal-checkout');
 
+var ASYNC_DEPENDENCY_TIMEOUT = 30000;
+
 function BasePayPalView() {
   BaseView.apply(this, arguments);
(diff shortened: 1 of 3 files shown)


src/views/payment-sheet-views/base-paypal-view.js
```diff
@@ -21,7 +21,10 @@ BasePayPalView.prototype._initialize = function (isCredit) {
 
   this.model.asyncDependencyStarting();
   asyncDependencyTimeoutHandler = setTimeout(function () {
-    self.model.asyncDependencyFailed({view: self.ID});
+    self.model.asyncDependencyFailed({
+      view: self.ID,
+      error: {message: 'There was an error connecting to PayPal.'}
```
crookedneighbor: Can we use `DropinError` instead?

PR #165, opened 2017-05-13
Add ability to pass in a DOM Node to Drop-in creation

### LINKED ISSUE #159: Container input support
I was skimming some of the source code in lieu of documentation (see #158), when I noticed this: https://github.com/braintree/braintree-web-drop-in/blob/182bd39/src/dropin.js#L47

```
  if (!this._merchantConfiguration.selector) {
    analytics.sendEvent(this._client, 'configuration-error');
    callback(new DropinError('options.selector is required.'));
    return;
  }

  container = document.querySelector(this._merchantConfiguration.selector);
```

In Braintree v2, directly passing in `container` with a DOM element was supported (and my existing code depends on this). Would you be willing to add that option back in this module?

crookedneighbor:
### Summary

Some frameworks don't give you easy access to css selectors for the dom nodes where drop-in will be inserted. This allows people to pass in a `container` property instead of selector which will be the DOM Node to pass Drop-in into.

We _could_ overload `selector` and allow it to take a DOM Node or a CSS selector, but I thought it would be a weird API for prop named selector to take a DOM Node directly.

Closes #159 

### Checklist

- [x] Added a changelog entry
- [x] Ran unit tests


src/dropin.js
```diff
@@ -37,26 +37,38 @@ Dropin.prototype = Object.create(EventEmitter.prototype, {
...
 
   this._injectStylesheet();
 
-  if (!this._merchantConfiguration.selector) {
+  if (!selector && !container) {
     analytics.sendEvent(this._client, 'configuration-error');
-    callback(new DropinError('options.selector is required.'));
+    callback(new DropinError('options.selector or options.container are required.'));
+    return;
+  } else if (selector && container) {
```
lilaconlee: Is this better than picking one by default? I could go either way.

PR #166, opened 2017-05-17
Add updateConfig method to dropin

### LINKED ISSUE #87: Dynamic PP amount?

crookedneighbor:
Closes #87

### Summary

Provides a way for merchants to modify their PayPal settings without tearing down Drop-in and re-creating it.

This can be done before PayPal authentication without issue.

If a customer authenticates their PayPal account and then `updateConfig` is called, the authenticated PayPal account will be removed. In addition, if they are on the options `methods` view, it'll return to the `options` view.

### Checklist

- [x] Added a changelog entry
- [x] Ran unit tests
- [ ] Verify that it works in the vault flow with client token generated with a customer id
- [x] Verify that it works with a merchant that has only paypal enabled
- [ ] Add JSDocs


src/dropin.js
```diff
@@ -9,6 +9,8 @@ var EventEmitter = require('./lib/event-emitter');
 var isGuestCheckout = require('./lib/is-guest-checkout');
 var fs = require('fs');
 var MainView = require('./views/main-view');
+var PaymentMethodsView = require('./views/payment-methods-view');
```
lilaconlee: Can we just require the IDs here?

src/dropin.js
```diff
@@ -133,6 +135,32 @@ Dropin.prototype._initialize = function (callback) {
   }.bind(this));
 };
 
+Dropin.prototype.updateConfig = function (prop, key, value) {
+  var methodsViewId = PaymentMethodsView.ID;
+
+  if (prop !== 'paypal' && prop !== 'paypalCredit') {
```
lilaconlee: I could see us making a whitelist in the constants just to make it abundantly clear what this is doing, but could be  overkill at this point.

src/dropin.js
```diff
@@ -133,6 +143,36 @@ Dropin.prototype._initialize = function (callback) {
...
+Dropin.prototype.updateConfiguration = function (prop, key, value) {
+  var methodsViewId = PaymentMethodsView.ID;
+
+  if (UPDATABLE_CONFIGURATION_OPTIONS.indexOf(prop) === -1) {
+    return;
+  }
+
+  this._mainView.getView(prop).updateConfiguration(key, value);
+
+  if (UPDATABLE_CONFIGURATION_OPTIONS_THAT_REQUIRE_DELETION.indexOf(prop) === -1) {
```
lilaconlee: Trying to think of a better name for this...

src/dropin.js
```diff
@@ -252,7 +292,8 @@ function formatPaymentMethodPayload(paymentMethod) {
   var formattedPaymentMethod = {
     nonce: paymentMethod.nonce,
     details: paymentMethod.details,
-    type: paymentMethod.type
+    type: paymentMethod.type,
+    vaulted: true
```
lilaconlee: This is the payload we return to merchants, we should make sure there isn't a downside to this and we've removed all references to the payment method object having the same shape as the v3 tokenization payloads.

PR #170, opened 2017-05-20
Readme.md: Reference the release branch for localization

ClashTheBunny:
### Summary

The commit 9657cc8cbf95233791d60de7ecf8b66d5e326795 updated the readme to point to the release branch of the docs until the JSDocs are available.  This fixes the reference to the localization link until that point also.  All other links look like they are functioning.  This would also need to be updated when the JSDocs are ready.

### Checklist

- [ ] Added a changelog entry
- [ ] Ran unit tests

I did not do either of those because this is a documentation change.  It would appear that there are not unit tests for the documentation.  It would also be strange to document this in the changelog as other larger documentation changes have not been.


crookedneighbor:
:+1:

Issue #175, opened 2017-05-23
JSDoc task causes build to fail on windows


juongithub:
### General information

* SDK/Library version: unreleased (after 1.0.2)
* OS: Windows 8.1 64-bit

### Issue description

When I run 'npm run build' I see the errors:
```
[15:20:26] Using gulpfile [PATH]\braintree-web-drop-in\gulpfile.js

[15:20:26] Starting 'build'...
[15:20:26] Starting 'clean'...
[15:20:26] Finished 'clean' after 93 ms
[15:20:26] Starting 'build:js:unmin'...
[15:20:26] Starting 'build:js:min'...
[15:20:26] Starting 'build:css'...
[15:20:26] Starting 'build:demoapp'...
[15:20:26] Starting 'build:npm:statics'...
[15:20:26] Starting 'build:npm:package.json'...
[15:20:26] Starting 'build:npm:src'...
[15:20:26] Finished 'build:npm:package.json' after 66 ms
[15:20:27] Finished 'build:demoapp' after 670 ms
[15:20:27] Starting 'build:gh-pages'...
[15:20:27] Starting 'jsdoc:generate'...
The system cannot find the file specified.
[15:20:27] 'jsdoc:generate' errored after 59 ms
[15:20:27] Error: 1
    at formatError ([PATH]\braintree-web-drop-in\node_modules\gulp\bin\gulp.js:169:10)
    at Gulp.<anonymous> ([PATH]\braintree-web-drop-in\node_modules\gulp\bin\gulp.js:195:15)
    at emitOne (events.js:101:20)
    at Gulp.emit (events.js:188:7)
    at Gulp.Orchestrator._emitTaskDone ([PATH]\braintree-web-drop-in\node_modules\orchestrator\index.js:264:8)
    at [PATH]\braintree-web-drop-in\node_modules\orchestrator\index.js:275:23
    at finish ([PATH]\braintree-web-drop-in\node_modules\orchestrator\lib\runTask.js:21:8)
    at cb ([PATH]\braintree-web-drop-in\node_modules\orchestrator\lib\runTask.js:29:3)
    at ChildProcess.<anonymous> ([PATH]\braintree-web-drop-in\gulpfile.js:204:7)
    at emitTwo (events.js:106:13)
[15:20:27] 'build' errored after 921 ms
[15:20:27] Error in plugin 'run-sequence(jsdoc:generate)'
Error
[...]

PR #176, opened 2017-05-23
'sed' is not available on Windows (issue #175)

juongithub:
JSDoc task causes build to fail on Windows since 'sed' is not available on this platform. The function _replaceVersionInFile create a new file and delete it after the jsdoc is executed.


crookedneighbor:
Made some modifications here: https://github.com/braintree/braintree-web-drop-in/pull/179

Can you confirm that you're able to build with that PR?

PR #179, opened 2017-05-23
Allow JSDocs to be built on Windows system

crookedneighbor:
### Summary
Includes changes from https://github.com/braintree/braintree-web-drop-in/pull/176

The main difference is that I moved the code to interpolate the version to _after_ the jsdocs are built, so there's no need to have a temporary file.

### Checklist

- [ ] ~~Ran unit tests~~

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -158,8 +158,10 @@ gulp.task('build', function (done) {
   done);
 });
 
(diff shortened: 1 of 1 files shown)


juongithub:
The build process works correctly.

PR #188, opened 2017-05-26
Script tag integration

crookedneighbor:
### Summary

Adds script tag integration to Drop-in.

screen shot 2017-05-26 at 11 24 08 am

### Checklist

- [x] Added a changelog entry
- [x] Ran unit tests

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,9 +1,9 @@
 'use strict';
 /**
  * @module braintree-web-drop-in
- * @description There are two ways to integrate Drop-in into your page. You can use [`dropin.create`](#.create) or a standalone script tag integration (example below).
+ * @description The primary way to integrate Drop-in into your page is to use [`dropin.create`](#.create).
  *
- * If you only need to process credit cards on your checkout form, the script tag integration is the simplest and easiest way to integrate. All you need to do is add the Dropin script inside your form where you want Drop-in to appear and include a `data-braintree-dropin-authorization` property with your [tokenization key](https://developers.braintreepayments.com/guides/authorization/tokenization-key/javascript/v3) or [client token](https://developers.braintreepayments.com/guides/authorization/client-token).
+ * Alternatively, if you only need to process credit cards on your checkout page, the script tag integration is the simplest way to integrate. All you need to do is add the Drop-in script inside your form element where you want Drop-in to appear and include a `data-braintree-dropin-authorization` property with your [tokenization key](https://developers.braintreepayments.com/guides/authorization/tokenization-key/javascript/v3) or [client token](https://developers.braintreepayments.com/guides/authorization/client-token).
  *
  * The script tag integration will intercept the form submission and attempt to tokenize the credit card. If the tokenization is successful, it will insert the payment method nonce representing the credit card into a hidden input with the name `payment_method_nonce` and then submit your form.
  *
--- a/src/index.js
+++ b/src/index.js
@@ -1,10 +1,35 @@
 'use strict';
 /**
  * @module braintree-web-drop-in
+ * @description There are two ways to integrate Drop-in into your page. You can use [`dropin.create`](#.create) or a standalone script tag integration (example below).
+ *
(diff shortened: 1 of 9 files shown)


src/index.js
```diff
@@ -1,10 +1,35 @@
 'use strict';
 /**
  * @module braintree-web-drop-in
+ * @description There are two ways to integrate Drop-in into your page. You can use [`dropin.create`](#.create) or a standalone script tag integration (example below).
```
crookedneighbor: Is this where this belongs? Should it be mentioned on the home page of the docs as well?

src/index.js
```diff
@@ -225,6 +250,9 @@ function setAnalyticsIntegration(clientInstance) {
   return clientInstance;
 }
 
+// we check for document's existence to support server side rendering
+createFromScriptTag(create, document && document.querySelector('script[data-braintree-dropin-authorization]'));
```
crookedneighbor: Since we have to look for this data attribute, I wanted to make sure it was unique. Is this overboard? Should it just be `[data-braintree-authorization]` or `[data-dropin-authorization]` or something else?

If we didn't support IE, we could just do `document.currentScript` :/

EvanHahn: This will give a `ReferenceError` in server-side rendering. I believe the expression should be:

```js
typeof document !== 'undefined' && document.querySelector('script[data-braintree-dropin-authorization]')
```

I could see throwing a `DropinError` if it finds multiple scripts.

src/index.js
lilaconlee: I think we could mention it briefly on the homepage as well. I could see us emphasizing `dropin.create` as the main way to integrate with the script tag as an alternative, where this kinda reads like the script tag integration is the default.

src/index.js
```diff
@@ -1,10 +1,35 @@
 'use strict';
 /**
  * @module braintree-web-drop-in
+ * @description There are two ways to integrate Drop-in into your page. You can use [`dropin.create`](#.create) or a standalone script tag integration (example below).
+ *
+ * If you only need to process credit cards on your checkout form, the script tag integration is the simplest and easiest way to integrate. All you need to do is add the Dropin script inside your form where you want Drop-in to appear and include a `data-braintree-dropin-authorization` property with your [tokenization key](https://developers.braintreepayments.com/guides/authorization/tokenization-key/javascript/v3) or [client token](https://developers.braintreepayments.com/guides/authorization/client-token).
```
lilaconlee: Maybe we can clarify what "on your checkout form" means here. "As part of your checkout `form` element"? That doesn't sound quite right either but I think we need to emphasize that it needs to be in a form.

Also don't know if we want to say "Dropin script" or "Drop-in script" here.

src/lib/create-from-script-tag.js
```diff
@@ -0,0 +1,65 @@
...
+function createFromScriptTag(createFunction, scriptTag) {
+  var authorization, containerId, container, form;
+
+  if (!scriptTag) {
+    return;
+  }
+
+  authorization = scriptTag.getAttribute('data-braintree-dropin-authorization');
+
+  if (!authorization) {
```
lilaconlee: Do we want to throw an error if there is a script tag but no authorization?

src/lib/create-from-script-tag.js
```diff
@@ -0,0 +1,65 @@
...
+      throw createError;
+    }
+
+    form.addEventListener('submit', function (event) {
+      event.preventDefault();
+
+      instance.requestPaymentMethod(function (requestPaymentError, payload) {
+        var paymentMethodNonce = form.querySelector('[name="payment_method_nonce"]');
+
+        if (requestPaymentError) {
```
lilaconlee: We could grab the nonce after checking for the error here.

test/app/script-tag-integration.html
```diff
@@ -0,0 +1,49 @@
+<!doctype html>
```
lilaconlee: Do we want to include this in the hosted demo app? Don't see much of advantage there but we could do it.

src/lib/create-from-script-tag.js
```diff
@@ -0,0 +1,65 @@
...
+
+  if (!form) {
+    throw new DropinError('No form found for script tag integration.');
+  }
+
+  form.insertBefore(container, scriptTag);
+
+  createFunction({
+    authorization: authorization,
+    selector: '#' + containerId
```
lilaconlee: I'd use `container` here

test/app/script-tag-integration.html
crookedneighbor: If we don't change anything, it will be automatically added when we deploy it next.

PR #190, opened 2017-05-30
Add custom translations

crookedneighbor:
### Summary

Allows merchants to pass in an object to `translations` for custom translations. Any translation strings that are not provided will use the provided locale or fallback to the `en_US` version.

### Checklist

- [x] Added a changelog entry
- [x] Ran unit tests

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -20,9 +20,9 @@ var VERSION = process.env.npm_package_version;
  * @param {string} options.authorization A [tokenization key](https://developers.braintreepayments.com/guides/authorization/tokenization-key/javascript/v3) or a [client token](https://developers.braintreepayments.com/guides/authorization/client-token). If authorization is a client token created with a [customer ID](https://developers.braintreepayments.com/guides/drop-in/javascript/v3#customer-id), Drop-in will render saved payment methods and automatically store any newly-added payment methods in their Vault record.
  * @param {string|HTMLElement} options.container A reference to an empty element, such as a `<div>`, where Drop-in will be included on your page or the selector for the empty element. e.g. `#dropin-container`.
  * @param {string} options.selector Deprecated: Now an alias for `options.container`.
- * @param {string|object} [options.locale=`en_US`] Use this option to change the language, links, and terminology used throughout Drop-in. Supported locales include:
- * `de_DE`,
+ * @param {string} [options.locale=`en_US`] Use this option to change the language, links, and terminology used throughout Drop-in. Supported locales include:
  * `da_DK`,
+ * `de_DE`,
  * `en_AU`,
  * `en_GB`,
  * `en_US`,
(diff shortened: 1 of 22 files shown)


src/index.js
```diff
@@ -20,12 +20,12 @@ var VERSION = process.env.npm_package_version;
  * @param {string} options.authorization A [tokenization key](https://developers.braintreepayments.com/guides/authorization/tokenization-key/javascript/v3) or a [client token](https://developers.braintreepayments.com/guides/authorization/client-token). If authorization is a client token created with a [customer ID](https://developers.braintreepayments.com/guides/drop-in/javascript/v3#customer-id), Drop-in will render saved payment methods and automatically store any newly-added payment methods in their Vault record.
  * @param {string|HTMLElement} options.container A reference to an empty element, such as a `<div>`, where Drop-in will be included on your page or the selector for the empty element. e.g. `#dropin-container`.
  * @param {string} options.selector Deprecated: Now an alias for `options.container`.
- * @param {string} [options.locale=`en_US`] Use this option to change the language, links, and terminology used throughout Drop-in. Supported locales include:
- * `da_DK`,
+ * @param {string|object} [options.locale=`en_US`] Use this option to change the language, links, and terminology used throughout Drop-in. Supported locales include:
  * `de_DE`,
- * `en_US`,
+ * `da_DK`,
```
quinnjn: Should this be alphabetical?

skunkworks:
The implementation looks good from what I can tell, but it feels funky to have `locale` be a string or object. What are your thoughts on adding a separate `translations` or `localizationStrings` parameter that takes an object? It would take precedence over `locale` if defined.

crookedneighbor:
That would alleviate some of the wackiness in the paypal view.

src/index.js
```diff
@@ -171,6 +173,18 @@ var VERSION = process.env.npm_package_version;
...
  *   </body>
  * </html>
+ *
+ * @example
+ * <caption>Use your own translations</caption>
+ * braintree.dropin.create({
+ *   authorization: 'CLIENT_AUTHORIZATION',
+ *   container: '#dropin-container',
+ *   locale: {
+ *     payingWith: 'You are paying with {{paymentSource}}',
```
quinnjn: This is slightly confusing to me. It looks like the merchant has to specify the strings they want for the customer rather than providing Drop-in with unsupported locales.

Say I want to support multiple locales, but I notice that Drop-in does not support `en_CA`. I might want to add my own `en_CA` but I still want Drop-in to load in my customer locales (`fr_CA`, `en_US`, etc).

Would it be better to have:
```
localizationOverride: {
    en_CA: {
      payingWith: '...'
    }
  }
```

src/index.js
crookedneighbor: Whoops. Yeah, it should be.

intelliot:
What wackiness in the paypal view?

skunkworks:
@intelliot The type-checking of `locale` to see if it's a locale or an object with string translations

PR #199, opened 2017-06-07
Prevent default in script tag integration before Drop-in is created

lilaconlee:
### Summary
Adds an event listener with `event.preventDefault` before Drop-in is created with a script tag integration so the form can't be submitted before the form appears.

Open to more elegant solutions. Possible issue could be if Drop-in fails to load, we'll still be preventing default on the form.

### Checklist
~Added a changelog entry~
- [x] Ran unit tests


lilaconlee:
I was thinking of merging this into master then merging master into the promisify branch in case we want to a release before promisification

PR #201, opened 2017-06-08
Add locale support to script tag integration

crookedneighbor:
### Summary

Adds a `data-locale` attribute to script tag integration.

Not sure what the best way to document this stuff is. Maybe it should have it's own page in JSDoc?

### Checklist

- [x] Added a changelog entry
- [x] Ran unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 CHANGELOG
 =========
 
+unreleased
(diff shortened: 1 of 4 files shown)


crookedneighbor:
Closing this in favor of a more generic solution so we don't need to keep opening up new prs for new config options

PR #209, opened 2017-06-19
Fix error when using Hosted Fields overrides to remove CVV

lilaconlee:
### Summary

Fixes a bug where errors in the console appeared when removing the CVV field with Hosted Fields overrides. We were trying to update the cvv placeholder and such even when that field did not exist.

Diff has a lot of moves because I moved the CVV challenges logic into the `_generateHostedFieldsOptions` method.

### Checklist

- [x] Added a changelog entry
- [x] Ran unit tests

### DIFF
--- a/src/views/payment-sheet-views/card-view.js
+++ b/src/views/payment-sheet-views/card-view.js
@@ -121,46 +121,46 @@ CardView.prototype._generateHostedFieldsOptions = function () {
     }
   };
 
-  if (overrides) {
-    if (overrides.fields) {
-      if (overrides.fields.cvv && overrides.fields.cvv.placeholder) {
-        this._hasCustomCVVPlaceholder = true;
-      }
-
-      Object.keys(overrides.fields).forEach(function (field) {
-        if ((field === 'cvv' || field === 'postalCode') && overrides.fields[field] === null) {
-          delete options.fields[field];
-          return;
-        }
-
-        if (!options.fields[field]) {
-          return;
-        }
-
-        assign(options.fields[field], overrides.fields[field], {
-          selector: options.fields[field].selector
-        });
-      });
-    }
-
-    if (overrides.styles) {
-      Object.keys(overrides.styles).forEach(function (style) {
-        if (overrides.styles[style] === null) {
-          delete options.styles[style];
-          return;
-        }
-
-        assign(options.styles[style], overrides.styles[style]);
-      });
-    }
-  }
-
   if (!hasCVVChallenge) {
     delete options.fields.cvv;
   }
 
   if (!hasPostalCodeChallenge) {
     delete options.fields.postalCode;
+  }
+
+  if (!overrides) { return options; }
+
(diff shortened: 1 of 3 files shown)


src/views/payment-sheet-views/card-view.js
```diff
@@ -25,33 +25,27 @@ CardView.ID = CardView.prototype.ID = constants.paymentOptionIDs.card;
...
+  this.hasCVV = hfOptions.fields.cvv;
   this.cardNumberIcon = this.getElementById('card-number-icon');
   this.cardNumberIconSvg = this.getElementById('card-number-icon-svg');
   this.cvvIcon = this.getElementById('cvv-icon');
   this.cvvIconSvg = this.getElementById('cvv-icon-svg');
   this.cvvLabelDescriptor = this.getElementById('cvv-label-descriptor');
   this.fieldErrors = {};
 
-  if (!hasCVV || !hfOptions.fields.cvv) {
+  if (!hfOptions.fields.cvv) {
```
crookedneighbor: This is identical to `!this.hasCVV`, right?

src/views/payment-sheet-views/card-view.js
```diff
@@ -124,40 +121,46 @@ CardView.prototype._generateHostedFieldsOptions = function () {
...
+    }
   }
 
-  if (overrides.styles) {
-    Object.keys(overrides.styles).forEach(function (style) {
-      if (overrides.styles[style] === null) {
-        delete options.styles[style];
-        return;
-      }
+  if (!hasCVVChallenge) {
```
crookedneighbor: We probably want people to be able to supply the cvv if it's specified in the overrides (in case they're verifying on the server).

So this could be:

```js
var hasCVVOverride;
var hasPostalCodeOverride;

// stuff

if (overrides) {
  if (overrides.fields) {
    hasCVVOverride = Boolean(overrides.fields.cvv);
    hasPostalCodeOverride = Boolean(overrides.fields.postalCode);

// stuff

if (!hasCVVChallenge && !hasCVVOverride) {
```

lilaconlee: Talked in real life, we're not going to do this as part of this PR

src/views/payment-sheet-views/card-view.js
```diff
@@ -77,6 +71,9 @@ CardView.prototype._initialize = function () {
 };
 
 CardView.prototype._generateHostedFieldsOptions = function () {
```
skunkworks: I'm having some difficulty understanding what this method does now. Not sure if it's because of the `if` nesting, or because it has a lot of special cases. Is it possible to move the special cases to the top and do an early return? Would it be an improvement to add each field option based on whitelisted conditions instead of initializing options and removing them based on blacklist conditions?

These are just suggestions, not demands :)

lilaconlee: I'll give another go at cleaning it up!

PR #213, opened 2017-06-22
Add payment option chosen event

### LINKED ISSUE #212: Braintree option changed event
### General information

* SDK/Library version: https://js.braintreegateway.com/web/dropin/1.3.1/js/dropin.min.js
* Environment: Sandbox
* Browser and Chrome 59 (64-bit) on Win10

### Issue description

Hi,

I feel selecting a braintree option should also emit an event. I'm talking about these:
image

The only correct way to tie the drop-in into for example an AngularJS app is via events. Everything else means simply hacking your way around the limitations. And in my opinion the payment process shouldn't be a half-baked hacked up solution. It's great that the drop-in lib exists, but many apps nowadays are SPA's and these kind of events would be of great help.

Thanks!

crookedneighbor:
closes #212

### Summary

Adds event for when a payment option is chosen.

```js
dropinInstance.on('paymentOptionSelected', function (event) {
  console.log(event.paymentOption, 'selected.'); // card, paypal, paypalCredit
});
```

### Checklist

- [x] Added a changelog entry
- [x] Ran unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ unreleased
 - `dropinInstance.requestPaymentMethod` will return a promise if no callback is provided
 - `dropinInstance.teardown` will return a promise if no callback is provided
 - `dropin.create` will return a promise if no callback is provided
+- Add `paymentOptionChosen` event
 
 1.3.1
 -----
--- a/test/unit/dropin.js
+++ b/test/unit/dropin.js
@@ -1028,6 +1028,22 @@ describe('Dropin', function () {
     });
   });
 
+  describe('payemnt option selected event', function () {
+    it('emits paymentOptionSelected when the model emits paymentOptionSelected', function (done) {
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ unreleased
 - `dropinInstance.requestPaymentMethod` will return a promise if no callback is provided
 - `dropinInstance.teardown` will return a promise if no callback is provided
 - `dropin.create` will return a promise if no callback is provided
+- Add `paymentOptionChosen` event
```
lilaconlee: Think we just need to update this to be `paymentOptionSelected`

test/unit/dropin.js
```diff
@@ -1028,6 +1028,22 @@ describe('Dropin', function () {
     });
   });
 
+  describe('payemnt option selected event', function () {
```
skunkworks: `payment`

PR #215, opened 2017-06-22
Fix error where drop-in would not finish transition if inside a hidden container

### LINKED ISSUE #210: "Choose another way to pay" unclickable after payment
### General information

* 1.3.1
* Sandbox
* Chrome 57.0.2987.98 (64-bit)  and Firefox 53.0.2 (64-bit)

### Issue description

As a user with already one saved credit card I want the page to reload after a payment goes through the server, regardless of success or failure, and to be able to make another payment using the drop-in UI. I want to be able to choose from the saved credit or  'choose another way to pay'.

After a payment goes through on the server the page refreshes. The user has a list of items that he can click pay on which will pop open a modal with the drop-in UI, and the majority of the time the "Choose another way to pay" link/button will not be clickable. I am assuming there is some sort of caching on braintree's side because the few times it works it is because the whole drop-in component reloads.

crookedneighbor:
closes #210

### Summary

The transition end listener never fires if the drop-in is inside a hidden container. This means that some parts of the drop-in are obscured by an invisible loader div that never gets cleaned up.

This adds a simple check if the element containing drop-in, or a parent is hidden and skips the transition listener.

### Checklist

- [x] Added a changelog entry
- [x] Ran unit tests

### DIFF
--- a/src/lib/transition-helper.js
+++ b/src/lib/transition-helper.js
@@ -2,8 +2,20 @@
 
 var browserDetection = require('./browser-detection');
 
+function isHidden(element) {
+  if (!element) { // no parentNode, so nothing containing the element is hidden
+    return false;
+  }
+
+  if (element.style.display === 'none') {
+    return true;
+  }
+
+  return isHidden(element.parentNode);
+}
+
 function onTransitionEnd(element, propertyName, callback) {
-  if (browserDetection.isIe9()) {
+  if (browserDetection.isIe9() || isHidden(element)) {
     callback();
     return;
   }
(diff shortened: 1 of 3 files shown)


src/lib/transition-helper.js
```diff
@@ -2,8 +2,20 @@
 
 var browserDetection = require('./browser-detection');
 
+function isHidden(element) {
+  if (!element) { // no parentNode, so nothing containing the element is hidden
+    return false;
+  }
+
+  if (element.style.display === 'none') {
```
crookedneighbor: May also need to check visibility prop here.

crookedneighbor: Confirmed that `visibility` and `opacity` do not affect it.

PR #219, opened 2017-06-29
Smooth payment methods view animation

lilaconlee:
### Summary
When immediately submitting the form after receiving a nonce, you can briefly see the payment methods view which is a little weird. Adding a little `setTimeout` there seems to take care of it without interrupting the flow that ends in the payment methods view.

If anyone has ideas for a more elegant solution I'm definitely open, would much prefer something not involving `setTimeout`. 

### Checklist

- [x] Added a changelog entry
- [x] Ran unit tests


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ unreleased
 - `dropin.create` will return a promise if no callback is provided
 - Fix error thrown in console when removing fields with card overrides
 - Fix bug where Drop-in would not finish loading if inside a hidden div
+- Fix transition from payment sheet views to payment methods view
```
quinnjn: It isn't really a fix, more of an improvement but 🤷‍♂️

test/unit/views/main-view.js
```diff
@@ -570,6 +573,20 @@ describe('MainView', function () {
       };
     });
 
+    describe('for changeActivePaymentMethod', function () {
```
quinnjn: Could drop `for`

PR #223, opened 2017-07-05
Add optional selectedPaymentMethod when emitting paymentMethodRequestable

### LINKED ISSUE #211: Selected method changed event? paymentMethodRequestable inconsistency
### General information

* SDK/Library version: https://js.braintreegateway.com/web/dropin/1.3.1/js/dropin.min.js
* Environment: Sandbox, AngularJS
* Browser and Chrome 59 (64-bit) on Win10

### Issue description

Hi all,

I feel that having `paymentMethodRequestable` and `noPaymentMethodRequestable` is not really enough to get a flawless workflow with the drop-in. Let's take these few images as an example of what I'm thinking of:

Step 1.
 - user enters the details of a card he wants to use
 - `paymentMethodRequestable` fired, it's now safe for the user to click the "Select payment method" button so we unhide it
 - the user clicks the button button which calls `dropInInstance.requestPaymentMethod()`
 - we obtain the nonce, hide the "Select payment method" button, the drop-in changes to "Paying with card" view and we display the "Pay" button - Step 2.
 - if the user clicks the "Pay" button, we make a request to the back-end which then creates a transaction
 - all fine so far

payment - quickdbd - google chrome

Step 2.
 - the user changes his mind, and clicks the "Choose another way to pay" button
 - at this point `noPaymentMethodRequestable` event has fired 
 - we hide both the "Select payment" and "Pay" button because it makes sense to do so
 - the user is taken to Step 3.

payment - quickdbd - google chrome_2

Step 3.
 - this is where the inconsistency lies and where I feel the problem is

Step 3.1.
 - if we click on the top button (original card details ending with 11), `paymentMethodRequestable` is fired but the drop-in also automatically takes you back to the Step 2., only this time with the wrong button being displayed ("Select payment method")
 - this time we don't have to click that button because the drop-in is already aware of the current nonce, our app has it cached in current controller instance and making an additional request in unnecessary
 - having to click this button one more time is redundant and brings a bit of confusion to the UX
 - instead, we want to display the "Pay" button

Step 3.2.
 - if we enter new card details the `paymentMethodRequestable` fires after the last detail is entered
 - now we can display the "Select payment method" button which in this case is correct
 - we click the button, it calls the `dropInInstance.requestPaymentMethod()`, gets the new nonce and the user gets taken to Step 2.
 - this time the correct button is displayed ("Pay")

payment - quickdbd - google chrome_3

To me it seems that with the current combinations of events which the drop-in lib provides, it's impossible to get the workflow to avoid that redundant "Select payment method" click. Am I right or is there perhaps a way to do it and I'm totally clueless? Is there a way to know whether the `paymentMethodRequestable` came from clicking on an existing payment method or from filling the form with the last bit of information?

I believe the needed information (origin of event) could be put inside the `even` param of the `paymentMethodRequestable` event callback. That would make these two events distinguishable and it would help a lot.

Thanks!

quinnjn:
### Summary
When emitting `paymentMethodRequestable` events we can now optionally send the `selectedPaymentMethod` back as well. 

This is optionally sent back because some `paymentMethodRequestable` events do not yet have the payment method such as when the card form is in a valid state.

Fixes #211 

### Checklist

- [X] Added a changelog entry
- [X] Ran unit tests

### DIFF
--- a/src/dropin-model.js
+++ b/src/dropin-model.js
@@ -79,14 +79,11 @@ DropinModel.prototype._shouldEmitRequestableEvent = function (options) {
 DropinModel.prototype.setPaymentMethodRequestable = function (options) {
   var shouldEmitEvent = this._shouldEmitRequestableEvent(options);
   var paymentMethodRequestableResponse = {
+    isPaymentMethodSelected: Boolean(options.selectedPaymentMethod),
     type: options.type
   };
 
   this._paymentMethodIsRequestable = options.isRequestable;
-
-  if (options.selectedPaymentMethod) {
-    paymentMethodRequestableResponse.selectedPaymentMethod = options.selectedPaymentMethod;
-  }
 
   if (options.isRequestable) {
     this._paymentMethodRequestableType = options.type;
--- a/src/dropin-model.js
+++ b/src/dropin-model.js
@@ -78,8 +78,15 @@ DropinModel.prototype._shouldEmitRequestableEvent = function (options) {
 
 DropinModel.prototype.setPaymentMethodRequestable = function (options) {
   var shouldEmitEvent = this._shouldEmitRequestableEvent(options);
+  var paymentMethodRequestableResponse = {
+    type: options.type
+  };
 
   this._paymentMethodIsRequestable = options.isRequestable;
+
+  if (options.selectedPaymentMethod) {
+    paymentMethodRequestableResponse.selectedPaymentMethod = options.selectedPaymentMethod;
+  }
 
   if (options.isRequestable) {
     this._paymentMethodRequestableType = options.type;
@@ -92,7 +99,7 @@ DropinModel.prototype.setPaymentMethodRequestable = function (options) {
   }
 
   if (options.isRequestable) {
-    this._emit('paymentMethodRequestable', {type: options.type});
+    this._emit('paymentMethodRequestable', paymentMethodRequestableResponse);
   } else {
     this._emit('noPaymentMethodRequestable');
   }
--- a/src/dropin-model.js
+++ b/src/dropin-model.js
@@ -79,7 +79,7 @@ DropinModel.prototype._shouldEmitRequestableEvent = function (options) {
 DropinModel.prototype.setPaymentMethodRequestable = function (options) {
   var shouldEmitEvent = this._shouldEmitRequestableEvent(options);
   var paymentMethodRequestableResponse = {
-    isPaymentMethodSelected: Boolean(options.selectedPaymentMethod),
+    paymentMethodIsSelected: Boolean(options.selectedPaymentMethod),
     type: options.type
   };
 
--- a/src/views/main-view.js
+++ b/src/views/main-view.js
@@ -170,7 +170,8 @@ MainView.prototype.setPrimaryView = function (id, secondaryViewId) {
 
   this.model.setPaymentMethodRequestable({
     isRequestable: Boolean(paymentMethod),
-    type: paymentMethod && paymentMethod.type
+    type: paymentMethod && paymentMethod.type,
+    selectedPaymentMethod: paymentMethod
   });
 
   this.model.clearError();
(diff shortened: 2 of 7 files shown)


crookedneighbor:
Got mixed feelings about this. Part of me thinks that calling `requestPaymentMethod` should be the source of truth for getting a payment method object and it seems strange to include it here, but I understand the use case.

pootzko:
@crookedneighbor I agree with what you're saying. Indeed it doesn't feel right for these events to pass the selectedPaymentMethod. I know I'm not part of the team or anything, but as a user of your library, I'll still leave my opinion. :D I hope that's ok. :)

I think the root of the problem is that you have two very simple events `paymentMethodRequestable` and `noPaymentMethodRequestable` for what should probably be a single event (obviously if there are any deeper reasons for doing it this way, I might not be aware of them). To me personally it feels like it's a bit of an overkill to have a separate events for these two. For example if you had a form with a radio-button field with 4 options, you wouldn't make 4 different events for when each one of these is clicked. You would have a single `changed` event which would then pass the currently selected option when fired. This seems to be kind of the same situation.

So ultimately, what I feel would make most sense in this case would be to have a `paymentMethodChanged` which would always pass the `selectedPaymentMethod` which would describe the currently selected option. And that method would among other things have a boolean property `isRequestable` so the user can decide what to do with that information further depending on what he needs/wants.

So that might be the way to go, it's really your decision. But even these changes you made will help me a lot. :)
Much appreciated! Thanks.

crookedneighbor:
I've changed `selectedPaymentMethod` to a boolean called `isPaymentMethodSelected` and included examples in the documentation for how to utilize it with `requestPaymentMethod`.

src/dropin-model.js
```diff
@@ -78,6 +78,10 @@ DropinModel.prototype._shouldEmitRequestableEvent = function (options) {
 
 DropinModel.prototype.setPaymentMethodRequestable = function (options) {
   var shouldEmitEvent = this._shouldEmitRequestableEvent(options);
+  var paymentMethodRequestableResponse = {
+    isPaymentMethodSelected: Boolean(options.selectedPaymentMethod),
```
crookedneighbor: I have a slight preference for `paymentMethodIsSelected` so that the if block reads:

```js
if (event.paymentMethodIsSelected) {
```

@intelliot do you have thoughts about that?

src/views/main-view.js
```diff
@@ -170,7 +170,8 @@ MainView.prototype.setPrimaryView = function (id, secondaryViewId) {
 
   this.model.setPaymentMethodRequestable({
     isRequestable: Boolean(paymentMethod),
-    type: paymentMethod && paymentMethod.type
+    type: paymentMethod && paymentMethod.type,
+    selectedPaymentMethod: paymentMethod
```
crookedneighbor: I elected to send along the `selectedPaymentMethod` here, rather than doing the `isPaymentMethodSelected` logic here. Is that 👍  with everyone?

src/dropin-model.js
intelliot: I'm fine with `paymentMethodIsSelected`

simeyla:
Can someone explain why `paymentMethodRequestable` provides me with the payment method, but `isPaymentMethodRequestable` doesn't?

If you need to know the payment method, then surely you need to know it whether it's preselected or explicitly been clicked? 

Also I **really** need to know when payment section is *incomplete*. For instance if somebody clicks 'PayPal' but then doesn't click on the actual express checkout button. It's way too presumptious to assume people will click on it. When I then request the method I get `{"name":"DropinError","message":"No payment method is available."}` so I can't even tell the user 'Please be sure to click the PayPal button'.

crookedneighbor:
@simeyla 	I've already addressed this in the other issue you opened up, but we're tracking your request in https://github.com/braintree/braintree-web-drop-in/issues/433

PR #225, opened 2017-07-07
Test shared dependencies are the same version

quinnjn:
### Summary
We should always use the same shared dependency versions. This check will help us verify if our versions are off.

The publishing tests are separated from our normal tests because they are only useful before a release, and we may update dependency versions between releases.

### Checklist

- [ ] Added a changelog entry
- [X] Ran unit tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "pretest": "npm run lint",
     "test": "karma start test/config/karma.js --single-run",
     "test:integration": ". ./.env && bundle exec rake sauce:spec test_files=spec",
-    "test:publishing": "npm run pretest && ./node_modules/mocha/bin/mocha test/publishing",
+    "test:publishing": "npm run pretest && mocha test/publishing",
     "deploy:gh-pages": "./scripts/deploy-gh-pages"
   },
   "repository": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -10,6 +10,7 @@
     "pretest": "npm run lint",
     "test": "karma start test/config/karma.js --single-run",
     "test:integration": ". ./.env && bundle exec rake sauce:spec test_files=spec",
+    "test:publishing": "npm run pretest && ./node_modules/mocha/bin/mocha test/publishing",
```
crookedneighbor: You shouldn't need to do `./node_modules/mocha/bin/mocha`. `mocha` should use the node modules version when run from an npm script.

test/publishing/dependency-check.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+var dependencies = require('../../package.json').dependencies;
+var dependencyNames = Object.keys(dependencies);
+var packages = dependencyNames.filter(function (d) {
+  // Remove scoped packages
+  return d[0] !== '@';
```
crookedneighbor: Why do we take out the scoped package? That's one of the core shared dependencies.

intelliot:
This looks pretty good. We may want to add something like this to braintree-web to catch cases like https://github.com/braintree/braintree-web/issues/297

PR #226, opened 2017-07-07
Fix svgs in Firefox when inside a base element

### LINKED ISSUE #224: Broken icons in Firefox with Angular routing
### General information

* SDK/Library version: 1.3.1
* Browser and OS: Firefox 54.0.1 (64-Bit) on Windows 10

### Issue description

Firefox has a problem displaying SVGs with `<use>` tags when there is a `<base href="/">` element present. This base tag is required in Angular applications for routing to work properly.

The easiest way to reproduce this issue is to use Firefox to open https://braintree.github.io/braintree-web-drop-in/ and use the dev tools to add `<base href="/">` to the HTML head. Using the teardown and create buttons, the issue can be observed because the icons are broken.

crookedneighbor:
closes #224

### Summary

Sets the values of the `xlink-href` to the page url + the hash of the svg.

### Checklist

- [x] Added a changelog entry
- [x] Ran unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 CHANGELOG
 =========
 
+unreleased
+----------
+- Fix bug where Firefox would not render svgs when inside of a base element
+
 1.4.0
 ------
 - Add `paymentOptionSelected` event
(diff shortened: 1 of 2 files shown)


crookedneighbor:
Closing in favor of providing the workaround in the drop-in create call until Firefox releases a fixed version in August.

https://github.com/braintree/braintree-web-drop-in/issues/224#issuecomment-314137191

PR #237, opened 2017-07-24
Cardholder name

crookedneighbor:
### Summary

Adds cardholder name to card field.

* Can take a `boolean` (or, more accurately, a truthy value) or an `object` to configure. If a boolean, it will set `required` to false.
* If required is set to true, client side validation will kick in to prevent tokenization until the field is not empty

### Checklist

- [x] Added a changelog entry


src/translations/en_US.js
```diff
@@ -25,12 +26,14 @@ module.exports = {
...
   unsupportedCardTypeError: 'This card type is not supported. Please try another card.',
   // Card form
+  cardholderNameLabel: 'Cardholder Name',
   cardNumberLabel: 'Card Number',
   cvvLabel: 'CVV',
   cvvThreeDigitLabelSubheading: '(3 digits)',
   cvvFourDigitLabelSubheading: '(4 digits)',
   expirationDateLabel: 'Expiration Date',
   expirationDateLabelSubheading: '(MM/YY)',
+  cardholderNamePlaceholder: 'The Name on your Card',
```
lilaconlee: We should run this by the designers before we get translations, maybe we don't have a placeholder here?

src/views/payment-sheet-views/card-view.js
```diff
@@ -70,6 +79,44 @@ CardView.prototype._initialize = function () {
...
+      classlist.remove(cardholderNameField, 'braintree-form__field-group--has-error');
+    }
+
+    this._sendRequestableEvent();
+  }.bind(this), false);
+
+  if (cardholderNameOptions.required) {
+    this.cardholderNameInput.addEventListener('blur', function () {
+      if (this.cardholderNameInput.value.length === 0) {
+        classlist.add(cardholderNameField, 'braintree-form__field-group--has-error');
```
lilaconlee: We should change this to only show the error when another field is focused like we do for Hosted Fields

test/unit/views/payment-sheet-views/card-view.js
```diff
@@ -192,6 +192,36 @@ describe('CardView', function () {
...
+        mainView: this.mainView,
+        model: this.model,
+        client: this.client,
+        strings: strings
+      });
+
+      expect(this.element.querySelector('[data-braintree-id="cardholder-name-field-group"]')).to.exist;
+    });
+
+    it('doesnot include cardholderName if not provided in merchant configuration', function () {
```
lilaconlee: does not

test/unit/views/payment-sheet-views/card-view.js
```diff
@@ -192,6 +192,36 @@ describe('CardView', function () {
       expect(this.element.querySelector('[data-braintree-id="postal-code-field-group"]')).not.to.exist;
     });
 
+    it('has cardholderName if provided in merchant configuration', function () {
```
lilaconlee: should we add tests around whether or not cardholder name is required and hiding/showing errors

crookedneighbor: Yes, will add.

PR #239, opened 2017-07-24
Move publishing test to separate travis build

crookedneighbor:
### Summary

NPM has some weird behavior with `prepublish` scripts in that they run whenever you `npm install`. So this test is actually being run before each build (along with linting). 

This PR moves it to a separate build on Travis.

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,9 +12,8 @@
     "test:integration": ". ./.env && bundle exec rake sauce:spec test_files=spec",
     "test:integration:paypal-only": "RUN_PAYPAL_ONLY=true npm run test:integration",
     "test:integration:paypal-skipped": "SKIP_PAYPAL=true npm run test:integration",
-    "test:publishing": "npm run pretest && mocha test/publishing",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -12,9 +12,8 @@
     "test:integration": ". ./.env && bundle exec rake sauce:spec test_files=spec",
     "test:integration:paypal-only": "RUN_PAYPAL_ONLY=true npm run test:integration",
     "test:integration:paypal-skipped": "SKIP_PAYPAL=true npm run test:integration",
-    "test:publishing": "npm run pretest && mocha test/publishing",
-    "deploy:gh-pages": "./scripts/deploy-gh-pages",
-    "prepublish": "npm run test:publishing"
+    "test:publishing": "mocha test/publishing",
```
intelliot: we don't need to lint anymore?

skunkworks: I think I'm OK with linting only occurring during `npm run test`. @crookedneighbor?

crookedneighbor: Yes, that's it precisely. Linting happens as a pretest action.

PR #240, opened 2017-07-26
Clear active payment method

crookedneighbor:
### Summary

Adds a `clearActivePaymentMethod` method which removes the selected payment method when called.

The use case is that there are some cases where the saved nonce can be consumed, but drop-in has no idea that it is consumed and shows a UI to the customer that the nonce can be submitted.

For instance, if you have a single page app and you send the nonce to your server and the transaction fails, you can update your ui to instruct the customer to try another payment method, but drop-in is stuck showing the ✅ .

This method removes that when called.

Another use case will be for 3DS, which consumes the nonce. IE, if liability does not shift, the merchant may want the customer to try a different payment method.

If this method is called on the vaulted payment method view, it leaves the vaulted payment methods present, but removes the ✅  from the selected one (and hides the `Paying with <TYPE>` message.

### Checklist

- [x] Added a changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ unreleased
 - Update browser-detection to v1.5.0
 - Add `aria-label` attribute to payment options
 - Update checkout.js to v4.0.95
-- Add `clearActivePaymentMethod` to remove selected payment method
+- Add `clearSelectedPaymentMethod` to remove selected payment method
 
 1.4.0
 ------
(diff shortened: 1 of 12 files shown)


crookedneighbor:
Is `clearActivePaymentMethod` a clear name? I was also toying with `clearSelectedPaymentMethod`.

crookedneighbor:
Publishing test worked! Updated browser detection.

Issue #229, opened 2017-07-12
Number keypad on mobile does not hide on form submit


danieljodeci:
I am having an issue on mobile devices where the number keypad is not hiding after initiating the requestPaymentMethod request on the click event. I'm using the example for v3.

crookedneighbor:
@danieljodeci Can you provide specific devices where this is occurring so we can reproduce the behavior?

crookedneighbor:
Was able to reproduce this with an iOS device. Working on a fix. Thanks for the report!

danieljodeci:
Hey! Sorry for the delay, but yes I was experiencing this issue on an iPhone 6. Thanks for the quick response!

crookedneighbor:
This will be fixed in the next release.

crookedneighbor:
This is now fixed in v1.5.0

PR #241, opened 2017-07-26
Update braintre-web to v3.20.1

crookedneighbor:
### Summary
Updates braintree-web to v3.20.1

I expect the publishing build to fail (just want to do a sanity check that it is). Will update browser detection when it does.

### Checklist

- [x] Added a changelog entry


crookedneighbor:
Closes #229

PR #243, opened 2017-07-31
Use jsdoc-template 3.1.0

lilaconlee:
### Summary

Includes anchor tags on examples and wider max-width for tables


Button0501:
Ok

PR #252, opened 2017-08-09
Data colllector

crookedneighbor:
### Summary

Adds data collector to drop-in.

### Checklist

- [x] Added a changelog entry

### DIFF
--- a/src/dropin.js
+++ b/src/dropin.js
@@ -183,7 +183,7 @@ Dropin.prototype = Object.create(EventEmitter.prototype, {
 });
 
 Dropin.prototype._initialize = function (callback) {
-  var localizedStrings, localizedHTML, strings;
+  var localizedStrings, localizedHTML;
   var dropinInstance = this; // eslint-disable-line consistent-this
   var container = this._merchantConfiguration.container || this._merchantConfiguration.selector;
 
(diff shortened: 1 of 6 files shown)


src/dropin.js
```diff
@@ -357,6 +351,33 @@ Dropin.prototype.clearSelectedPaymentMethod = function () {
...
 
+Dropin.prototype._setUpDataCollector = function () {
+  var config = assign({}, this._merchantConfiguration.dataCollector, {client: this._client});
+
+  this._model.asyncDependencyStarting();
+
+  dataCollector.create(config).then(function (instance) {
+    this._deviceData = instance.deviceData;
+  }.bind(this)).catch(function (err) {
+    console.log('Data Collector failed to set up', err);
```
crookedneighbor: Not sure what the right thing to do here is. I've elected to log to the console that data collector couldn't be set up and continue on as if the dependency didn't fail (since there's no UI updates to make to the Drop-in for this failure).

Open to suggestions for a better way to handle this.

src/index.js
```diff
@@ -172,6 +179,9 @@ var VERSION = process.env.npm_package_version;
  * @param {object} [options.paypalCredit] The configuration options for PayPal Credit. To include a PayPal Credit option in your Drop-in integration, include the `paypalCredit` parameter and [enable PayPal in the Braintree Control Panel](https://developers.braintreepayments.com/guides/paypal/testing-go-live/#go-live).
  *
  * Some of the PayPal Credit configuration options are listed [here](#~paypalCreateOptions), but for a full list see the [PayPal Checkout client reference options](http://braintree.github.io/braintree-web/{@pkg bt-web-version}/PayPalCheckout.html#createPayment). For more information on PayPal Credit, see the [Braintree Developer Docs](https://developers.braintreepayments.com/guides/paypal/paypal-credit/javascript/v3).
+ *
```
lilaconlee: Should we document the data collector failure behavior here?

src/index.js
```diff
@@ -120,6 +120,13 @@ var VERSION = process.env.npm_package_version;
  * @param {object} [overrides.styles] The Hosted Fields [`styles` options](http://braintree.github.io/braintree-web/{@pkg bt-web-version}/module-braintree-web_hosted-fields.html#~styleOptions).
  */
 
+/**
+ * @typedef {object} dataCollectorOptions The configuration options for Data Collector. Requires [advanced fraud protection](https://developers.braintreepayments.com/guides/advanced-fraud-tools/client-side/javascript/v3) to be enabled in the Braintree gateway. Contact our [support team](mailto:support@braintreepayments.com) to configure your Kount ID.
```
lilaconlee: We could use the support contact form instead of email here https://developers.braintreepayments.com/forms/contact

src/index.js
crookedneighbor: Agreed. We should update the data collector docs too.

src/index.js
crookedneighbor: 3714 496353 98431

PR #262, opened 2017-08-15
Refactor view initialization

crookedneighbor:
### Summary

We call `_initialize` within the constructor for the payment sheet views. This causes side effects beyond instantiating a view.

Instead, we should call `initialize` after creating the view and allow `initialize` to return a promise so we can test it easier.

(this will also make adding unionpay support easier)

PayPal should also be converted, but I have not completed the work yet.

### Checklist

- [ ] ~~Added a changelog entry~~ (internal change only)

### DIFF
--- a/src/views/main-view.js
+++ b/src/views/main-view.js
@@ -62,6 +62,7 @@ MainView.prototype._initialize = function () {
         client: this.client,
         strings: this.strings
       });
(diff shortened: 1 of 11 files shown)


crookedneighbor:
Added a commit to promisify the paypal view to match.

PR #267, opened 2017-08-24
Document binData and standardize requestPaymentMethod payload

lilaconlee:
### Summary
Previously, we were only formatting vaulted payment methods. This formats all payment methods before returning them in `requestPaymentMethod`. In the future we'll have to add new params to be returned.

We also document `binData`.
### Checklist

~[ ] Added a changelog entry~

### DIFF
--- a/src/dropin.js
+++ b/src/dropin.js
@@ -40,6 +40,7 @@ var VERSION = process.env.npm_package_version;
  * @property {string} details.lastTwo Last two digits of card number.
  * @property {string} description A human-readable description.
  * @property {string} type The payment method type, always `CreditCard` when the method requested is a card.
+ * @property {object} binData Information about the card based on the bin. Documented {@link Dropin~binData|here}.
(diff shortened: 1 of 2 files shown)


src/dropin.js
```diff
@@ -501,12 +515,8 @@ Dropin.prototype._disableErroredPaymentMethods = function () {
  */
 Dropin.prototype.requestPaymentMethod = function () {
   return this._mainView.requestPaymentMethod().then(function (payload) {
-    if (this._dataCollectorInstance) {
-      payload.deviceData = this._dataCollectorInstance.deviceData;
-    }
-
-    return payload;
-  }.bind(this));
+    return formatPaymentMethodPayload(payload);
```
crookedneighbor: I don't see how the deviceData is being added now, since it never gets added to the payload.

lilaconlee: Yeeeeah not what we meant to do there. Good catch.

PR #271, opened 2017-08-31
Allow useraction to be set for PayPal button.

plevetsky:
### Summary
`paypal.button.render()` in PayPal's checkout.js accepts a `commit` parameter to control whether the customer sees a `Pay Now` or `Continue` experience on the PayPal review page.  By default, the `Continue` experience is shown, but it is often preferable to show the `Buy Now` experience.

When the final payment amount is known at the time the PayPal Checkout flow is initiated, and no further customer interaction is required before payment is attempted, the `Pay Now` experience should be used ([PayPal Best Practice](https://developer.paypal.com/docs/integration/direct/express-checkout/integration-jsv4/best-practices/#implement-the-pay-now-experience)).

Switching to the `Pay Now` experience also results in the payment amount being shown on the PayPal review page.

### Checklist

- [x] Added a changelog entry

### DIFF
--- a/src/views/payment-sheet-views/base-paypal-view.js
+++ b/src/views/payment-sheet-views/base-paypal-view.js
@@ -44,7 +44,7 @@ BasePayPalView.prototype.initialize = function () {
     checkoutJSConfiguration = {
       env: environment,
       style: self.paypalConfiguration.buttonStyle || {},
-      commit: self.paypalConfiguration.commit || false,
+      commit: self.paypalConfiguration.commit,
       locale: locale,
       payment: function () {
         return paypalInstance.createPayment(self.paypalConfiguration).catch(reportError);
--- a/src/views/payment-sheet-views/base-paypal-view.js
+++ b/src/views/payment-sheet-views/base-paypal-view.js
@@ -44,6 +44,7 @@ BasePayPalView.prototype.initialize = function () {
     checkoutJSConfiguration = {
       env: environment,
       style: self.paypalConfiguration.buttonStyle || {},
+      commit: self.paypalConfiguration.commit || false,
       locale: locale,
       payment: function () {
         return paypalInstance.createPayment(self.paypalConfiguration).catch(reportError);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 CHANGELOG
 =========
 
+x.x.x
+------
+- Allow useraction to be set for PayPal button.
+
 1.7.0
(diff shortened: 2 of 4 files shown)


src/views/payment-sheet-views/base-paypal-view.js
```diff
@@ -44,6 +44,7 @@ BasePayPalView.prototype.initialize = function () {
     checkoutJSConfiguration = {
       env: environment,
       style: self.paypalConfiguration.buttonStyle || {},
+      commit: self.paypalConfiguration.commit || false,
```
crookedneighbor: I don't think we need to fall back to false here. If it's not present, it'll just be undefined.

haraldpdl:
👍 Should commit=true be followed through in paymentMethodRequestable to automatically submit the form? It does not make sense to show a PayPal Pay Now button if the customer has to click another pay now button to submit the checkout form.

haraldpdl:
A checkout pay now submit button is needed for card payments. If PayPal is chosen instead of a card payment and a Pay Now button (commit=true) is shown, the intention is to submit the form automatically and not have a need to separately click the checkouts own pay now button.

plevetsky:
Sure, totally agree. I was just thinking that if you pass `commit: true` you will know to submit the form in `paymentMethodRequestable` if `paymentOption == 'paypal'`.  It's only potentially ambiguous if you are using both variants in the same integration.

crookedneighbor:
@haraldpdl @plevetsky it should not automatically submit the form. 

If `commit: true` is passed in the config, the merchant already knows that and can listen on the [paymentMethodRequestable event](https://braintree.github.io/braintree-web-drop-in/docs/current/Dropin.html#event:paymentMethodRequestable), check if it's `paypal` and call `requestPaymentMethod` themselves.

There's no reason to make assumptions about how the merchant has set up their page. They may not be using a form at all.

Now, if we include `commit: true` as an option in the [script tag integration](https://braintree.github.io/braintree-web-drop-in/docs/current/module-braintree-web-drop-in.html) is another thing. In that case, it _may_ be correct to submit the form automatically when `commit: true` is passed.

haraldpdl:
@crookedneighbor yep - just confirming if it should be done in paymentMethodRequestable and meets the PayPal Best Practice guide.

@plevetsky noted that the total amount would be displayed. Hopefully this means line items can also be passed in the future.

Great work guys!

crookedneighbor:
Thanks! This will be available in the next release.

Issue #272, opened 2017-08-31
paymentOptionSelected not being emitted


kanedaki:
### General information

* SDK/Library version: 3
* Environment: sandbox
* Browser and OS:  Chrome Version 60.0.3112.113 (Official Build) (64-bit) on MacOS 10.11

### Issue description
```
        dropin
          .create({ authorization, container: '#braintree-button' })
          .then(dropinInstance => {
            dropinInstance.on('paymentOptionSelected', function() {
                console.log('paymentOptionSelected')
            })
            dropinInstance.on('paymentMethodRequestable', function() {
                console.log('paymentMethodRequestable')
            })
            dropinInstance.on('noPaymentMethodRequestable', function() {
              console.log('noPaymentMethodRequestable')
            })
            return dropinInstance
          })
```

I've a multiple payment methods dropin configured. Whenever I add a new payment method,
paymentMethodRequestable and noPaymentMethodRequestable, are being fired correctly, but when I change from one already created method to another, there is no event being fired.
![screen shot 2017-08-31 at 5 07 57 pm](https://user-images.githubusercontent.com/565038/29930575-faf7e99c-8e6e-11e7-8b2c-2b84e7ca04c6.png)
![screen shot 2017-08-31 at 5 08 38 pm](https://user-images.githubusercontent.com/565038/29930601-115f43ba-8e6f-11e7-8894-28f86040ea39.png)

In other words, when I click on the first option of the second image, and the green mark change from the second option to the first, no event is being fired. As far as I understood, `paymentOptionSelected` is fired when selecting a different option, right? If this is not the case, how can I detect a change?

Thanks in advance

kzurawel:
There is currently no way to detect changes in payment method within the same type of payment (e.g. from one credit card to another). The `paymenOptionSelected` event occurs when the user switches between _different types_ of payment (e.g. from credit card to PayPal). We will update our docs to make this more clear.

Would you mind telling us more about your use case? There might be another way to achieve what you're after.

Thanks!

crookedneighbor:
In addition, the paymentOptionSelected event only fires when a payment option is selected. Not when selecting a previously vaulted/stored payment method.

crookedneighbor:
We've updated the docs with #273 to reflect that. Going to close this now, but feel free to leave a comment if you have a use case for having an event when a payment method is selected.

kanedaki:
Hi,

[...]

PR #274, opened 2017-09-01
Allow vaulted payment methods to not be pre-selected on initialization

crookedneighbor:
### Summary

Related to https://github.com/braintree/braintree-web-drop-in/issues/272

If you have a submit-button-less page, there's no way to know if a customer has selected a saved payment method because the default payment method comes pre-selected. Since it's pre-selected, the payment method request-able event doesn't fire (because it was already requestable). 

By providing an option to turn off pre-selection, when the customer selects a saved payment method, the event does fire.

### Checklist

- [x] Added a changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ unreleased
 ----------
 - Simplify check for checkout.js on the merchant's page
 - Allow useraction to be set for PayPal button.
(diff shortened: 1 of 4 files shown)


crookedneighbor:
Closes #272

TMmcgillis:
Tessa McGillis

PR #275, opened 2017-09-06
Add hungarian translation

weeg:


lilaconlee:
Because Hungarian is not supported in our PayPal flows, we've opted to not support this as an official translation but you can supply your own translations when creating Drop-in. See our [reference documentation](https://braintree.github.io/braintree-web-drop-in/docs/current/module-braintree-web-drop-in.html#create-examples) for an example of passing a `translations` object into your `create` call.

weeg:
No problem, thanks for the info.

crookedneighbor:

Issue #283, opened 2017-09-27
missing validations on cardholderName


YamileAndrada:
### General information

* SDK/Library version: https://js.braintreegateway.com/web/dropin/1.8.0/js/dropin.min.js
* Environment: Sandbox
* Browser: chrome, mozilla,etc

### Issue description
missing validations on  cardholder name field
- I entered cards with names of 100 characters, I think that maximum will be 32, I'm not sure
- also I was able to save cards with special characters like 
#"$#"$#&%"$54w6$%(#('574738907

crookedneighbor:
The max character limit in the Braintree gateway is 255 characters for cardholder name. We should update the client side validations to account for that.

Characters such as `#"$#"$#&%"$54w6$%(#('574738907` are valid in the Braintree gateway. So they should be allowed via client side validation.

YamileAndrada:
Great! thanks for the information. I'll keep an eye for the client update with the restriction
Regards.

crookedneighbor:
Fixed in 1.9.0 https://github.com/braintree/braintree-web-drop-in/blob/master/CHANGELOG.md#changelog

PR #288, opened 2017-10-04
Limit cardholder name to 255 characters

kzurawel:
### Summary

This PR adds an additional validation check / error for cardholder name to ensure that it is less than 256 characters in length. Additionally, it refactors how we handle validation checks for "extra inputs".

### Checklist

- [x] Added a changelog entry


src/views/payment-sheet-views/card-view.js
```diff
@@ -49,10 +69,14 @@ CardView.prototype.initialize = function () {
     postalCodeFieldGroup.parentNode.removeChild(postalCodeFieldGroup);
   }
 
-  if (this.hasCardholderName) {
-    this._setupCardholderName(cardholderNameField);
-  } else {
-    cardholderNameField.parentNode.removeChild(cardholderNameField);
+  for (i = 0; i < this.extraInputs.length; i++) {
```
crookedneighbor: Is there a reason to use a for loop instead of forEach?

src/views/payment-sheet-views/card-view.js
```diff
@@ -356,26 +427,31 @@ CardView.prototype.showFieldError = function (field, errorMessage) {
...
-  this.hostedFieldsInstance.removeAttribute({
-    field: field,
-    attribute: 'aria-invalid'
-  });
+  if (!(input && isNormalFieldElement(input))) {
+    this.hostedFieldsInstance.removeAttribute({
+      field: field,
+      attribute: 'aria-invalid'
+    });
+  }
```
crookedneighbor: We probably want to set and remove aria-invalid on normal inputs as well.

src/views/payment-sheet-views/card-view.js
kzurawel: No, will update this

src/views/payment-sheet-views/card-view.js
kzurawel: 👍

crookedneighbor:
Closes #283

src/views/payment-sheet-views/card-view.js
```diff
@@ -263,13 +307,36 @@ CardView.prototype._validateForm = function (showFieldErrors) {
     }
   }
 
-  if (!this._validateCardholderName()) {
-    isValid = false;
+  if (this.extraInputs) {
+    for (i = 0; i < this.extraInputs.length; i++) {
```
crookedneighbor: This could just as easily be a forEach, right? Just use `return` in place of `continue`

crookedneighbor:
FYI: cardholder name integration tests are failing.

src/views/payment-sheet-views/card-view.js
```diff
@@ -516,12 +583,15 @@ CardView.prototype._isCardTypeSupported = function (cardType) {
   return supportedCardTypes.indexOf(configurationCardType) !== -1;
 };
 
+function isNormalFieldElement(element) {
```
lilaconlee: Is a `NormalFieldElement` any different than an `extraInput` in this case?

src/views/payment-sheet-views/card-view.js
```diff
@@ -74,40 +91,65 @@ CardView.prototype.initialize = function () {
...
+          });
         }
-      }.bind(this), 0);
-    }.bind(this), false);
+      }, 0);
+    }, false);
   }
 };
 
+CardView.prototype._removeExtraInput = function (extraInput) {
```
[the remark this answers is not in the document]
crookedneighbor: I think we're following the pattern of how the hosted fields stuff is set up, where we remove the postal code input if that challenge is not required.

I'd be fine with programmatically inserting it though. Should probably do the same (in a future pr) for the Hosted Fields inputs.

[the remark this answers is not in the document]
lilaconlee: Yeah I was just thinking because they're "extra" it's odd to include them by default, but I think it'd be the cleanest to just programatically generate all the fields.

kzurawel:
re-test this please

PR #290, opened 2017-10-11
Fix misnamed translations

lilaconlee:
### Summary

This fixes some translations that were misnamed and adds the scripts I used to fix them in to `scripts`.

I'd be fine not including the scripts if folks are opposed, I just figured they might be useful again sometime.

### Checklist

- [x] Added a changelog entry


scripts/check-missing-translations.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
```
crookedneighbor: Could we turn this into a test instead?

scripts/update-translation-key.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+var fs = require('fs');
```
crookedneighbor: Could we include a comment here that illustrates command line usage?

lilaconlee: Think we need that in addition to the bit in `DEVELOPMENT.md`?

crookedneighbor: Oh, I just didn't see that.

lilaconlee:
Moved to the missing translations script into a test, pinned until we get the translations that are actually missing not just misnamed.

PR #293, opened 2017-10-12
Add Apple Pay

skunkworks:
### Summary

Add Apple Pay to Drop-in.

### Open questions

- [ ] Locale customization and button type customization
  - Do we want to support the other button types?
  -  If we don't want to support other button types, we don't need locale because it's only relevant for non-`plain` button types (buy, donate, set-up)
- [ ] What happens when Apple Pay is not available in a region? Does drop-in display the error correctly?
 - [ ] Analytics events
   - Currently, there's only one Apple Pay analytics event to know when Apple Pay is enabled/available as a payment option. What do we want to know (if anything) about Apple Pay in Dropin that can't be queried via analytics in JS SDK?
     - `canMakePaymentsWithActiveCard` true/false?
       - Maybe we're interested to know how many people have Apple Pay but don't have card set up? Could add setup button flow.
     - How many Drop-in integrations have Apple Pay in supported vs. unsupported browsers?
     - Or more generally, how many Drop-in integrations have X payment option enabled/available?

### Checklist

- [x] Added a changelog entry
- [x] Add translations for strings


src/dropin-model.js
```diff
@@ -205,13 +206,27 @@ function getSupportedPaymentOptions(options) {
...
+    applePayBrowserSupported = window.ApplePaySession && ApplePaySession.canMakePayments(); // eslint-disable-line no-undef
+
+    if (!applePayEnabled) {
+      return false;
+    }
+    if (!applePayBrowserSupported) {
+      console.log('Browser does not support Apple Pay.'); // eslint-disable-line no-console
+      return false;
+    }
+    analytics.sendEvent(options.client, 'apple-pay.available');
```
crookedneighbor: I think this needs to be stubbed in the unit tests.

crookedneighbor: I've just pushed a commit that does this.

package.json
```diff
@@ -58,7 +58,8 @@
     "serve-static": "1.11.1",
     "sinon": "2.1.0",
     "sinon-chai": "2.9.0",
-    "vinyl-source-stream": "1.1.0"
+    "vinyl-source-stream": "1.1.0",
+    "watchify": "^3.9.0"
```
lilaconlee: did we intend to add watchify here?

src/dropin-model.js
```diff
@@ -205,13 +206,27 @@ function getSupportedPaymentOptions(options) {
...
     return gatewayConfiguration.paypalEnabled && Boolean(options.merchantConfiguration.paypalCredit);
+  } else if (paymentOption === 'applePay') {
+    applePayEnabled = gatewayConfiguration.applePay && Boolean(options.merchantConfiguration.applePay);
+    applePayBrowserSupported = window.ApplePaySession && ApplePaySession.canMakePayments(); // eslint-disable-line no-undef
+
+    if (!applePayEnabled) {
+      return false;
+    }
+    if (!applePayBrowserSupported) {
+      console.log('Browser does not support Apple Pay.'); // eslint-disable-line no-console
```
lilaconlee: Do we want to console log this or just document that Apple Pay will render in supported browsers in the docs?

src/index.js
```diff
@@ -140,6 +140,19 @@ var VERSION = process.env.npm_package_version;
  * @param {boolean} [commit] The user action to show on the PayPal review page. If true, a `Pay Now` button will be shown. If false, a `Continue` button will be shown.
  */
 
+/** @typedef {object} applePayCreateOptions The configuration options for Apple Pay.
+ *
+ * @param {string} [buttonStyle=`black`] Configures the Apple Pay button style. Valid values are `black`, `white`, `white-outline`. Default value is `black`.
+ * @param {string} displayName The canonical name for your store. Use a non-localized name. This parameter should be a UTF-8 string that is a maximum of 128 characters. The system may display this name to the user.
+ * @param {external:ApplePayPaymentRequest} paymentRequest The payment request details to apply on top of those from Braintree.
```
lilaconlee: Should we mention what details are provided by Braintree here?

src/translations/en_US.js
```diff
@@ -24,6 +24,11 @@ module.exports = {
   paypalAccountTokenizationFailedError: 'Something went wrong adding the PayPal account. Please try again.',
   paypalFlowFailedError: 'Something went wrong connecting to PayPal. Please try again.',
   paypalTokenizationRequestActiveError: 'PayPal payment authorization is already in progress.',
+  applePayAccountTokenizationFailedError: 'Something went wrong with Apple Pay. Please try again.',
+  applePayActiveCardError: 'Add a supported card to your Apple Pay wallet.',
+  applePayBrowserNotSupported: 'Browser does not support Apple Pay.',
```
lilaconlee: Do we want to show this error or hide Apple Pay as an option?

src/translations/en_US.js
```diff
@@ -24,6 +24,11 @@ module.exports = {
...
+  applePayFlowFailedError: 'Something went wrong initializing Apple Pay. Please try again.',
+  applePayTokenizationRequestActiveError: 'Apple Pay is already in progress.',
```
lilaconlee: Should it be "Apple Pay payment is already in progress..." or something like the PayPal error? Also can this error actually happen since the flow is part of the browser?

test/unit/views/payment-options-view.js
```diff
@@ -12,6 +12,12 @@ var analytics = require('../../../src/lib/analytics');
 var mainHTML = fs.readFileSync(__dirname + '/../../../src/html/main.html', 'utf8');
 
 var paymentOptionAttributes = {
+  applePay: {
+    icon: '#logoApplePay', // Get this
```
lilaconlee: We got this

src/translations/en_US.js
crookedneighbor: I think we just want it to not be shown. No need to have an error.

lilaconlee:
Thoughts on open questions:
* I'd say we add a ticket to support locales and alternate buttons as part of a separate PR, this one is bulky enough as it is.
* I'd say in all cases we know Apple Pay is not supported, we hide the payment method and do our best to document when it is and is not supported.
* In terms of analytics, we'll have all the events from the component, the event when the payment option is selected, and the event when an apple pay payment method is selected. Think it might make sense to add additional events in a separate PR just to make it super clear what we're adding, but I'm open to not doing that.

src/dropin-model.js
lilaconlee: Do we actually want this event? I think we get the same thing by looking at apple pay events with the `dropin2` integration type.

src/translations/en_US.js
lilaconlee: Looks like this error is not used so I'm going to go ahead and remove it.

src/translations/en_US.js
```diff
@@ -23,6 +23,11 @@ module.exports = {
   paypalAccountTokenizationFailedError: 'Something went wrong adding the PayPal account. Please try again.',
   paypalFlowFailedError: 'Something went wrong connecting to PayPal. Please try again.',
   paypalTokenizationRequestActiveError: 'PayPal payment authorization is already in progress.',
+  applePayAccountTokenizationFailedError: 'Something went wrong with Apple Pay. Please try again.',
```
lilaconlee: Didn't see this string used in the code and it isn't an error returned from bt.js

src/translations/en_US.js
crookedneighbor: This can be removed.

lilaconlee:
Think this PR is ready to go. We're going to tackle localization and JSDocs in a separate PR so it's a little easier to review.

src/index.js
skunkworks: I'm 👍 for that

lilaconlee: Think we're gonna beef up the jsdocs in a separate PR for easier reviewing

src/dropin-model.js
```diff
@@ -205,13 +205,20 @@ function getSupportedPaymentOptions(options) {
...
 
   if (paymentOption === 'card') {
     return gatewayConfiguration.creditCards.supportedCardTypes.length > 0;
   } else if (paymentOption === 'paypal') {
     return gatewayConfiguration.paypalEnabled && Boolean(options.merchantConfiguration.paypal);
   } else if (paymentOption === 'paypalCredit') {
     return gatewayConfiguration.paypalEnabled && Boolean(options.merchantConfiguration.paypalCredit);
+  } else if (paymentOption === 'applePay') {
+    applePayEnabled = gatewayConfiguration.applePay && Boolean(options.merchantConfiguration.applePay);
+    applePayBrowserSupported = window.ApplePaySession && ApplePaySession.canMakePayments(); // eslint-disable-line no-undef
```
crookedneighbor: We should use `global` instead of `window`. And if we use `global` for the `canMakePayment` function, we won't have to use `eslint-disable-line`

src/index.js
```diff
@@ -140,6 +140,19 @@ var VERSION = process.env.npm_package_version;
  * @param {boolean} [commit] The user action to show on the PayPal review page. If true, a `Pay Now` button will be shown. If false, a `Continue` button will be shown.
  */
 
+/** @typedef {object} applePayCreateOptions The configuration options for Apple Pay.
+ *
+ * @param {string} [buttonStyle=`black`] Configures the Apple Pay button style. Valid values are `black`, `white`, `white-outline`. Default value is `black`.
```
crookedneighbor: We shouldn't need to specify the default value, since that is indicated with `[buttonStyle=blank]`

src/views/payment-sheet-views/apple-pay-view.js
```diff
@@ -0,0 +1,91 @@
...
+ApplePayView.ID = ApplePayView.prototype.ID = paymentOptionIDs.applePay;
+
+ApplePayView.prototype.initialize = function () {
+  var self = this;
+
+  self.applePayConfiguration = assign({}, self.model.merchantConfiguration.applePay);
+
+  self.model.asyncDependencyStarting();
+
+  return btApplePay.create({client: this.client}).then(function (applePayInstance) { // eslint-disable-line consistent-return
```
crookedneighbor: We should investigate why eslint doesn't like this.

src/views/payment-sheet-views/apple-pay-view.js
```diff
@@ -0,0 +1,91 @@
...
+    var buttonDiv = self.getElementById('apple-pay-button');
+
+    self.applePayInstance = applePayInstance;
+
+    self.model.on('changeActivePaymentView', function (paymentViewID) {
+      if (paymentViewID !== self.ID) {
+        return;
+      }
+
+      ApplePaySession.canMakePaymentsWithActiveCard(self.applePayInstance.merchantIdentifier).then(function (canMakePayments) { // eslint-disable-line no-undef
```
crookedneighbor: Using `global.ApplePaySession` will fix this error.

src/views/payment-sheet-views/apple-pay-view.js
```diff
@@ -0,0 +1,91 @@
...
+      view: self.ID,
+      error: new DropinError(err)
+    });
+  });
+};
+
+ApplePayView.prototype._showPaymentSheet = function () {
+  var self = this;
+  var request = self.applePayInstance.createPaymentRequest(this.applePayConfiguration.paymentRequest);
+  var session = new ApplePaySession(2, request); // eslint-disable-line no-undef
```
crookedneighbor: Use `global` here.

src/views/payment-sheet-views/apple-pay-view.js
```diff
@@ -0,0 +1,91 @@
...
+      self.model.reportError(validationErr);
+      session.abort();
+    });
+  };
+
+  session.onpaymentauthorized = function (event) {
+    self.applePayInstance.tokenize({
+      token: event.payment.token
+    }).then(function (payload) {
+      session.completePayment(ApplePaySession.STATUS_SUCCESS); // eslint-disable-line no-undef
```
crookedneighbor: `global`

src/views/payment-sheet-views/apple-pay-view.js
```diff
@@ -0,0 +1,91 @@
...
+      payload.payment = event.payment;
+      self.model.addPaymentMethod(payload);
+    }).catch(function (tokenizeErr) {
+      self.model.reportError(tokenizeErr);
+      session.completePayment(ApplePaySession.STATUS_FAILURE); // eslint-disable-line no-undef
```
crookedneighbor: `global`

lilaconlee:
Ran these tests locally (and went through stuff manually) and everything is looking A+, gonna go ahead and merge.

PR #304, opened 2017-10-25
Fix issue where the edges of card form inputs were not clickable

crookedneighbor:
### Summary
Fix issue where the edges of card form inputs were not clickable

Applied this fix: https://github.com/braintree/braintree-web/issues/104#issuecomment-160668654

### Checklist

- [x] Added a changelog entry

### DIFF
--- a/src/html/main.html
+++ b/src/html/main.html
@@ -83,7 +83,7 @@
           <div data-braintree-id="number-field-group" class="braintree-form__field-group">
             <div class="braintree-form__label">{{cardNumberLabel}}</div>
             <div class="braintree-form__field">
-              <label for="braintree-form-number">
+              <label>
                 <div class="braintree-form-number braintree-form__hosted-field"></div>
               </label>
               <div class="braintree-form__icon-container">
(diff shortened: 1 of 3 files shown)


lilaconlee:
Are we worried about merchant styling of labels having odd behavior here?

crookedneighbor:
Yes!

src/html/main.html
```diff
@@ -83,7 +83,9 @@
           <div data-braintree-id="number-field-group" class="braintree-form__field-group">
             <div class="braintree-form__label">{{cardNumberLabel}}</div>
             <div class="braintree-form__field">
-              <div class="braintree-form-number braintree-form__hosted-field"></div>
+              <label for="braintree-form-number">
```
kzurawel: `for` takes an ID. Since we're including the input directly within the label (inside the div, anyways), I think we can leave it off here.

crookedneighbor: You are right, I thought we required it for having the label focus, but it looks like we just set it up to take label that surrounds the field as well: https://github.com/braintree/braintree-web/blob/b79008a51ccf3a002a20fb20a0df8976c39435d6/src/hosted-fields/external/hosted-fields.js#L515

PR #307, opened 2017-10-30
Upgrade firefox driver to latest version

crookedneighbor:
### Summary

The latest version of the Firefox driver does not have the same issue that previous versions had where you could not inspect the values of inputs inside iframes.

### Checklist


crookedneighbor:
Seems to be an issue with one of the update PP configuration tests with the latest version of the Firefox driver.

PR #325, opened 2017-11-20
Fix apple pay enabled logic

### LINKED ISSUE #324: Apple Pay not working in 1.9.0
### General information

* SDK/Library version: 1.9.0
* Environment: Sandbox
* Browser and OS: iOS 11

### Issue description

I've followed the instructions, and can't get Apple Pay working in the sandbox. Yes, I'm using a sandbox iCloud account, have a test credit card added, and `ApplePaySession.canMakePayments()` returns `true`.

From some debugging into the drop in, I'm seeing that my `gatewayConfiguration` is coming back with an `applePayWeb` field, but not `applePay` which is being looked for here:

https://github.com/braintree/braintree-web-drop-in/blob/2355dcbcdd733d5a8aa7a1420bc6160d56bddd27/src/dropin-model.js#L217

Is this a bug in the drop-in?

crookedneighbor:
### Summary

Closes #324

We accidentally had the logic set to check for `applePay` instead of `applePayWeb`. This worked because our test merchant had both iOS Apple Pay and web Apple Pay enabled.

Here's the check in the Apple Pay component: https://github.com/braintree/braintree-web/blob/0ae58971df381a59a7aa7b8eca7af378cbbaa19c/src/apple-pay/index.js#L30

### Checklist

- [x] Added a changelog entry

### DIFF
--- a/src/dropin-model.js
+++ b/src/dropin-model.js
@@ -214,7 +214,7 @@ function isPaymentOptionEnabled(paymentOption, options) {
   } else if (paymentOption === 'paypalCredit') {
     return gatewayConfiguration.paypalEnabled && Boolean(options.merchantConfiguration.paypalCredit);
   } else if (paymentOption === 'applePay') {
-    applePayEnabled = gatewayConfiguration.applePay && Boolean(options.merchantConfiguration.applePayWeb);
+    applePayEnabled = gatewayConfiguration.applePayWeb && Boolean(options.merchantConfiguration.applePay);
     applePayBrowserSupported = global.ApplePaySession && global.ApplePaySession.canMakePayments();
 
     if (!applePayEnabled || !applePayBrowserSupported) { return false; }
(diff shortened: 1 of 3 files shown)


src/dropin-model.js
```diff
@@ -214,7 +214,7 @@ function isPaymentOptionEnabled(paymentOption, options) {
   } else if (paymentOption === 'paypalCredit') {
     return gatewayConfiguration.paypalEnabled && Boolean(options.merchantConfiguration.paypalCredit);
   } else if (paymentOption === 'applePay') {
-    applePayEnabled = gatewayConfiguration.applePay && Boolean(options.merchantConfiguration.applePay);
+    applePayEnabled = gatewayConfiguration.applePay && Boolean(options.merchantConfiguration.applePayWeb);
```
dpolivy: I think the change is backwards -- in my testing, it was `gatewayConfiguration` that had `applePayWeb`, and I think `merchantConfiguration` did have `applePay`.

crookedneighbor: You're right, typo on my part. Fixing.

PR #328, opened 2017-11-21
Catch Apple Pay exception

maxsz:
### Summary

This catches an error that might happen when checking Apple Pay availablity. For example in development the ApplePaySession API might throw (InvalidAccessError (DOM Exception 15): Trying to call an ApplePaySession API from an insecure document) because the dropin is embedded in a non-https webpage (localhost).

### Checklist

- [x] Added a changelog entry

### DIFF
--- a/src/dropin-model.js
+++ b/src/dropin-model.js
@@ -215,7 +215,11 @@ function isPaymentOptionEnabled(paymentOption, options) {
     return gatewayConfiguration.paypalEnabled && Boolean(options.merchantConfiguration.paypalCredit);
   } else if (paymentOption === 'applePay') {
     applePayEnabled = gatewayConfiguration.applePayWeb && Boolean(options.merchantConfiguration.applePay);
-    applePayBrowserSupported = global.ApplePaySession && global.ApplePaySession.canMakePayments();
+    try {
+      applePayBrowserSupported = global.ApplePaySession && global.ApplePaySession.canMakePayments();
+    } catch (e) {
+      applePayBrowserSupported = false;
+    }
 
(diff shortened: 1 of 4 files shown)


src/dropin-model.js
```diff
@@ -215,7 +215,11 @@ function isPaymentOptionEnabled(paymentOption, options) {
     return gatewayConfiguration.paypalEnabled && Boolean(options.merchantConfiguration.paypalCredit);
   } else if (paymentOption === 'applePay') {
     applePayEnabled = gatewayConfiguration.applePayWeb && Boolean(options.merchantConfiguration.applePay);
-    applePayBrowserSupported = global.ApplePaySession && global.ApplePaySession.canMakePayments();
+    try {
```
crookedneighbor: Could we instead do:

```js
applePayBrowserSupported = global.ApplePaySession && global.location.protocol === 'https:' && global.ApplePaySession.canMakePayments();
```

maxsz: Sure. I've just tested it and it also works.

crookedneighbor:
Just needs a unit test and a changelog entry. Are you able to do that? Or do you need us to add them? Either way is fine, just let us know. 

Thanks for catching this!

maxsz:
I've added a basic unit test. But the other test fails, because `global.location.protocol` is always `http` during the test. Feel free to just fix it, if you have a solution :)

crookedneighbor:
:+1: Thanks for doing this. Should go out in the next release.

crookedneighbor:
This was released as part of 1.9.2. Thanks again!

PR #329, opened 2017-11-22
Fix for error when retrieving last two digits on saved Apple Pay payment method

julka:
### Summary

The `paymentInstrumentName` property isn't available on Apply Pay saved payment methods and trying to slice off the last two characters off undefined creates a JavaScript error. The result fro this is that the dropin payment form does not render for users with saved Apple Pay payment methods (when on a supported device).

Here's the error I'm seeing:

error

As you can see here, `this.paymentMethod.details` doesn't have a `paymentInstrumentName` property but it does have a `lastTwo` property. I think that's what you want to use here.

paymentmethod-object

### Checklist

- [x] Added a changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ CHANGELOG
 unreleased
 ----------
 - Improve logic for enabling Apple Pay to only trigger with HTTPS (#328 thanks @maxsz)
-- Fix error for saved ApplePay payment method (#332 thanks @julka)
+- Fix error for saved ApplePay payment method (#330 thanks @julka)
 
 1.9.1
 -----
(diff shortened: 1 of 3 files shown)


julka:
The previous commits created issues with new Apple Pay cards. 
image_uploaded_from_ios_720

It looks like there are two formats for `paymentMethod` with Apple Pay. One for new cards and one for saved cards. 

Updated code to work for both, put back old test (renamed) and added a test for the saved card format.

crookedneighbor:
Thanks for the PR. We'll look at this next week after the holidays. (same as with the other one 😄 )

crookedneighbor:
Thanks, I've simplified the logic a bit and merged this in. Should go out in the next release.

crookedneighbor:
This was released as part of 1.9.2. Thanks again!

PR #331, opened 2017-11-22
Shows Apple Pay on the small label for saved payment methods

julka:
### Summary

This code expects the Apple Pay saved payment method's `cardType` property to look something like `"Visa"` or `"Mastercard"` and thus match a string in one of your localization files, like here: https://github.com/braintree/braintree-web-drop-in/blob/master/src/translations/en_US.js#L49

The value in `cardType` actually looks like `"Apple Pay - Visa"` which is not defined in the localization files.

Here you can see the `cardType` property in `this.paymentMethod.details`:
paymentmethod-object

In this pull request I change this over to match the PayPal implementation, which hard codes to the property in the localization object for this payment method. 

And alternative fix would be to change this further upstream so that `cardType` for Apply Pay looks more like what it does with a standard credit card payment `cardType`.

### Checklist

- [x] Added a changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 CHANGELOG
 =========
 
+1.9.2
+-----
+- Displays "Apple Pay" instead of "undefined" for saved Apple Pay payment methods
+
 1.9.1
 -----
 - Normalize label styles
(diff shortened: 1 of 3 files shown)


julka:
I am happy to make changes to this PR with feedback. Let me know if there's anything I can do to move this along. 

Also, this PR and #329 will conflict. I'm happy to resolve the conflict in one after the other is merged.

crookedneighbor:
Thanks for the PR. We'll look at this next week after the holidays.

crookedneighbor:
Thanks for your work on this! Should go out in the next release.

crookedneighbor:
This was released as part of 1.9.2.

PR #343, opened 2018-01-05
Fix issue where vaulted apple pay methods were being displayed

crookedneighbor:
### Summary

Even if a vaulted apple pay method is presented on the client, it can't be used to run a transaction. This PR removes the vaulted apple pay methods from appearing as a choice for the customer and fixes a display issue where the wrong `ending in` text was being shown. Since there is no need to differentiate the Apple Pay method anymore, since it can't be vaulted, I've elected to include no text.

### Checklist

- [x] Added a changelog entry


djjames:
Thanks you

PR #349, opened 2018-01-22
Add kz_KZ to language support

Yerlan:


crookedneighbor:
Because Kazakh is not currently supported in our PayPal flows, we've opted to not support this as an official translation but you can supply your own translations when creating Drop-in. See our [reference documentation](https://braintree.github.io/braintree-web-drop-in/docs/current/module-braintree-web-drop-in.html#create-examples) for an example of passing a translations object into your create call.

PR #355, opened 2018-02-02
Update module dependencies

crookedneighbor:
### Summary

Updates bt web and promise-polyfill to fix a bug in promise-polyfill where IE browsers would throw errors.

### Checklist

- [x] Added a changelog entry


chhuang:
I'm still getting `'Promise' is undefined` after the update.

screen shot 2018-02-16 at 3 57 57 pm

PR #356, opened 2018-02-06
Update development instructions

crookedneighbor:
### Summary

This PR fixes up a couple of minor things and adds instructions for adding new payment methods to Drop-in.

### Checklist

- [ ] ~~Added a changelog entry~~ <- no bug fixes or public facing changes

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -54,7 +54,7 @@ When developing, you can include a locally built CSS file on the page that will 
 
 ## Adding new Payment Methods
 
-Adding a new payment method requires changing a number of files. For each of these sections, you will see an example of adding a fake payment method, `FooPay` to the file.
+Adding a new payment method requires changing a number of files. For each of these sections, you will see an example of adding a fake payment method, `FooPay` to the file. *Note:* This guide may fall out of date as things change in the repo. If you find a mistake, please open a PR to fix it.
 
 ### Constants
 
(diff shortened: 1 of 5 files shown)


DEVELOPMENT.md
```diff
@@ -52,6 +52,233 @@ When developing, you can include a locally built CSS file on the page that will
...
+
+```javascript
+'FooPay': 'FooPay'
+```
+
+### Drop-in Model
+
+The [src/dropin-model.js](https://github.com/braintree/braintree-web-drop-in/blob/master/src/dropin-model.js) must be updated so that Drop-in can check if the payment option is available for the customer to use.
+
+Primarilly, the `isPaymentOptionEnabled` function must be adjusted to account for the new payment method. It checks if the merchant is enabled for the particular payment method in the Braintree gateway, if the merchant has configured Drop-in to enable the payemnt method, and any other requirements the payment method may have to be used in Drop-in. For instance, if FooPay can only be used if the `FooPay` global exists on the window, we would probably add code that looks like this to `isPaymentOptionEnabled`.
```
skunkworks: `Primarily`

crookedneighbor:
Good idea @demerino

DEVELOPMENT.md
```diff
@@ -52,6 +52,233 @@ When developing, you can include a locally built CSS file on the page that will
...
+
+```javascript
+var DEFAULT_PAYMENT_OPTION_PRIORITY = [
+  // etc
+  paymentOptionIDs.applePay,
+  paymentOptionIDs.fooPay
+];
+```
+
+If a customer's vaulted payment methods cannot be used on the client (IE, ApplePay, GooglePay, Venmo), you must also add the payment method to the `VAULTED_PAYMENT_METHOD_TYPES_THAT_SHOULD_BE_HIDDEN` array. For instance:
```
quinnjn: > (IE, ApplePay, GooglePay, Venmo)

I _wish_ we could hide IE but I think this was supposed to be `i.e.,`:

DEVELOPMENT.md
```diff
@@ -52,6 +52,233 @@ When developing, you can include a locally built CSS file on the page that will
...
+```javascript
+var VAULTED_PAYMENT_METHOD_TYPES_THAT_SHOULD_BE_HIDDEN = [
+  paymentMethodTypes.ApplePayCard,
+  paymentMethodTypes.FooPayAccount
+];
+```
+
+### Main View
+
+The [src/html/main.html](https://github.com/braintree/braintree-web-drop-in/blob/master/src/html/main.html) must be updated with a div for the payment method's view. This is the UI where the payment method is initiated by the customer (putting in card details, pushing the PayPal button, etc). In our example, FooPay requires a button for the customer to press to initate the flow.
```
quinnjn: > putting in card details

filling the card form

DEVELOPMENT.md
```diff
@@ -52,6 +52,233 @@ When developing, you can include a locally built CSS file on the page that will
...
+  self.model.asyncDependencyReady();
+}).catch(function (err) {
+  self.model.asyncDependencyFailed({
+    view: self.ID,
+    error: new DropinError(err)
+  });
+});
+```
+
+If you don't need to handle specific errors, you can let Drop-in populate a generic error. If you do need to handle a specific error, you can either pass the key of a specific string to use in the translation file.
```
quinnjn: I feel like error convention could be its own header

PR #362, opened 2018-02-16
Fix non utf8 encoding handling

crookedneighbor:
### Summary

Merchants using sites that don't support utf-8 encoding get strange symbols for bullet points. This PR detects if the site is encoded as utf-8, and if not, swaps out the bullet points for `*`.

It also injects strings as html so that if a merchant is using a non-utf encoding, they can insert html encoded entities.

Lastly, we run some simple html sanitization against the merchant strings being passed in. This is not a huge concern, as anything a merchant might pass in via the translations they could just alter in Drop-in itself.

### Checklist

- [x] Added a changelog entry


src/lib/sanitize-html.js
```diff
@@ -0,0 +1,15 @@
...
+
+module.exports = function (string) {
+  if (typeof string !== 'string') {
+    return '';
+  }
+
+  return string
+    .replace(/</g, '&lt;')
+    .replace(/>/g, '&gt;');
+    // https://stackoverflow.com/a/1354491/2601552
```
jaypatel512: Any reason for not removing this ?

crookedneighbor: Whoops, I just left this in. Will definitely remove it.

PR #368, opened 2018-02-26
Add Google Pay to Drop-in

jaypatel512:
### Summary
- Add [Google Pay](https://developers.google.com/pay/api/) as a payment option for drop-in. 
- Created `assets` library that loads scripts and stylesheets.

#### Note:
To check if Google Pay is enabled, we needed to make an asynchronous call, that required us to refactor `DropinModel.isPaymentOptionEnabled` method. A static method `isEnabled` that returns a promise was added to each `PaymentSheetView` that verifies if the payment method is enabled or not.

### Checklist

- [x] Added a changelog entry


src/views/payment-sheet-views/google-pay-view.js
```diff
@@ -0,0 +1,117 @@
...
+
+  if (!(gatewayConfiguration.androidPay && Boolean(options.merchantConfiguration.googlePay))) {
+    return Promise.resolve(false);
+  }
+
+  return Promise.resolve().then(function () {
+    if (!global.google) {
+      return assets.loadScript(global.document.head, {
+        id: constants.GOOGLE_PAYMENT_SCRIPT_ID,
+        src: constants.GOOGLE_PAYMENT_SOURCE
```
demerino: Do we want to mention anywhere in the docs that Drop-in will load the google pay JS for you or is that not a common thing that we document?

src/dropin.js
```diff
@@ -226,17 +247,18 @@ Dropin.prototype = Object.create(EventEmitter.prototype, {
 
 Dropin.prototype._initialize = function (callback) {
   var localizedStrings, localizedHTML, paypalScriptOptions;
-  var dropinInstance = this; // eslint-disable-line consistent-this
-  var container = this._merchantConfiguration.container || this._merchantConfiguration.selector;
+  var self = this; // eslint-disable-line consistent-self
+  var container = self._merchantConfiguration.container || self._merchantConfiguration.selector;
```
quinnjn: I feel like we switch out `this` for `self` in a lot in this function. Is that required? I'd prefer to use `this` in most cases unless we have some crazy nested anonymous functions

src/dropin.js
```diff
@@ -246,110 +268,113 @@ Dropin.prototype._initialize = function (callback) {
...
 
-  this._dropinWrapper.innerHTML = svgHTML + localizedHTML;
-  container.appendChild(this._dropinWrapper);
+  self._dropinWrapper.innerHTML = svgHTML + localizedHTML;
+  container.appendChild(self._dropinWrapper);
 
-  this._getVaultedPaymentMethods().then(function (paymentMethods) {
-    var paypalRequired;
+  self._getVaultedPaymentMethods().then(function (paymentMethods) {
+    self._model = new DropinModel({
```
quinnjn: I'm OK with `self` in this scope.

src/dropin.js
```diff
@@ -246,110 +268,113 @@ Dropin.prototype._initialize = function (callback) {
...
-    }
+    return self._model.initialize();
+  }).then(function () {
+    var paypalRequired;
 
-    this._model.on('cancelInitialization', function (err) {
-      analytics.sendEvent(this._client, 'load-error');
-      this._dropinWrapper.innerHTML = '';
+    self._model.on('cancelInitialization', function (err) {
+      analytics.sendEvent(self._client, 'load-error');
```
quinnjn: This is a nitpick but I like reporting analytics just before we exit the function. So in this case, right before the `callback`

src/index.js
```diff
@@ -152,12 +152,24 @@ var VERSION = process.env.npm_package_version;
  * @param {external:ApplePayPaymentRequest} paymentRequest The payment request details to apply on top of those from Braintree.
  */
 
+/** @typedef {object} googlePayCreateOptions The configuration options for Google Pay. Additional options from the few listed here are available, many have default values applied based on the settings found in the Braintree Gateway. For more information, see [Google's Documentation](https://developers.google.com/pay/api/web/object-reference#request-objects).
+ *
+ * @param {string} merchantId The id provided by Google for processing transactions in production. Not necessary for testing in sandbox.
```
quinnjn: s/id/ID

If it is not necessary in sandbox would we want this to be optional in JSDoc?

test/unit/dropin-model.js
```diff
@@ -47,257 +53,255 @@ describe('DropinModel', function () {
...
-        this.sandbox.stub(venmo, 'isBrowserSupported').returns(false);
-
-        model = new DropinModel(this.modelOptions);
-
-        expect(model.supportedPaymentOptions).to.deep.equal(['card']);
-
-        venmo.isBrowserSupported.returns(true);
-
-        model = new DropinModel(this.modelOptions);
+    it('supports cards, PayPal, PayPal Credit, Apple Pay, Google Pay, and Venmo and defaults to showing them in correct paymentOptionPriority', function () {
```
quinnjn: Might want to name this test in the same priority order we expect

> supports cards, PayPal, PayPal Credit, Venmo, Apple Pay, and Google Pay

src/views/payment-sheet-views/google-pay-view.js
jaypatel512: I think we have the same setup for `PayPal`, and do not mention that in dropin. For dropin, basically we download all resources that we need ourselves.

src/index.js
jaypatel512: For JSDoc, I think keeping that as required makes sense, as the docs generally refer to the production environment.

src/dropin.js
crookedneighbor: I switched to using `self` everywhere in this function because there are my functions where we would have to bind(this) and it seemed cleaner to just use self everywhere. If you feel strongly about using this and bind for this function, we can, but I preferred going the self route.

quinnjn: This was more of a nitpick. `this` stands out as a keyword better than `self` but otherwise I'm OK with using `self`.

PR #369, opened 2018-02-26
Fix card view labels

### LINKED ISSUE #365: Labels do not focus fields
### General information

* SDK/Library version: 1.9.4
* Environment: Sandbox
* Browser and OS Chrome on Mac 10.13.2

### Issue description
Clicking on the label "Card number" does not properly focus the card number field.

Seems the hosted field example on the braintree support site works properly.

For now I am using the following workaround which seems to work in at least Chrome:

	$('.braintree-form__label').on('click', function () {
		$(this).next('.braintree-form__field').find('label').trigger('click');
	})

This is tricky to work around because the element needing focus is inside the protected iframe.

### LINKED ISSUE #367: Fields do not properly validate when focused through label
### General information

* SDK/Library version: 1.9.4
* Environment: Sandbox
* Browser and OS Chrome on mac 10.13.2

### Issue description

When focusing elements by clicking on a label, the validation does not trigger on blur which is inconsistent with when the fields are focused via keyboard (tab) or clicking directly on the field.

This is related to https://github.com/braintree/braintree-web-drop-in/issues/365

crookedneighbor:
### Summary

Allows card inputs to be selected by clicking their corresponding labels.

closes #365
closes #367

### Checklist

- [x] Added a changelog entry

### DIFF
--- a/src/views/payment-sheet-views/card-view.js
+++ b/src/views/payment-sheet-views/card-view.js
@@ -471,13 +471,18 @@ CardView.prototype._onBlurEvent = function (event) {
 
   classlist.remove(fieldGroup, 'braintree-form__field-group--is-focused');
 
-  if (isCardViewElement() && field.isEmpty) {
-    this.showFieldError(event.emittedBy, this.strings['fieldEmptyFor' + capitalize(event.emittedBy)]);
-  } else if (!field.isEmpty && !field.isValid) {
-    this.showFieldError(event.emittedBy, this.strings['fieldInvalidFor' + capitalize(event.emittedBy)]);
-  } else if (event.emittedBy === 'number' && !this._isCardTypeSupported(event.cards[0].type)) {
-    this.showFieldError('number', this.strings.unsupportedCardTypeError);
-  }
+  setTimeout(function () {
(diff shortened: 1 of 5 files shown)


src/views/payment-sheet-views/card-view.js
```diff
@@ -471,13 +471,22 @@ CardView.prototype._onBlurEvent = function (event) {
...
   }
+
+  setTimeout(function () {
+    // when focusing on a field by clicking the label,
+    // we need to wait a bit for the iframe to be
+    // focused properly before applying validations
+    if (shouldApplyFieldEmptyError(field)) {
+      this.showFieldError(event.emittedBy, this.strings['fieldEmptyFor' + capitalize(event.emittedBy)]);
+    }
+  }.bind(this), 150);
```
demerino: 💯 Hopefully this doesn't need to change.

jaypatel512: Yeah, we did a bit of manual testing to find that comfortable number. Generally, this is a very corner case, and even if the timeout is somehow inefficient, it would not be causing any major issue. There is a very tiny visual penalty involved if we go over that timeout period for any reason.

PR #373, opened 2018-03-05
Add option to persist card form data

crookedneighbor:
### Summary

Adds option to persist card data after tokenization. This can be useful for when the transaction fails and the customer needs to verify the card they input was correct.

### Checklist

- [x] Added a changelog entry


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ unreleased
 - Enable Google Pay support
 - Fix issue where non utf-8 encoded sites would show strange characters for card placeholders
 - Fix issue where card fields could not be focused by clicking on the corresponding label
+- Allow card form to not be cleared after succesful tokenization with `card.persistCardDataAfterTokenization`
```
crookedneighbor: How do we feel about the naming for this? Too verbose? Do we like the word `tokenization` in here? We don't use it in the Drop-in docs, even though that is what is happening.

idontusenumbers: Maybe something to the effect of 'allowEditingOfSavedPaymentMethods()' it's just as verbose but might be easier to understand.

crookedneighbor:
:+1: Except that we intentionally don't use the word tokenization in Drop-in. (hence why we have `requestPaymentMethod`)

crookedneighbor:
Also, just noticed that I used the word tokenization in my original version of this :shame:

crookedneighbor:
@quinnjn and I talked about it and we're just going to go with `clearCardFieldsAfterTokenization`

PR #377, opened 2018-03-20
Update duplicate card error handling to show better message

skunkworks:
### Summary

Improves messaging when the client token has been generated with `failOnDuplicatePaymentMethod` set to true.

### Checklist

- [ ] Get translations
- [ ] Added a changelog entry


src/translations/en_US.js
```diff
@@ -18,6 +18,7 @@ module.exports = {
   fieldInvalidForPostalCode: 'This postal code is not valid.',
   fieldTooLongForCardholderName: 'Cardholder name must be less than 256 characters.',
   genericError: 'Something went wrong on our end.',
+  hostedFieldsDuplicateCardError: 'This credit card already exists as a saved payment method.',
```
skunkworks: I'm open to changing the messaging around this to make it less awkward/technical sounding.

As a side note, I'm not sure if we need to improve our documentation around `failOnDuplicatePaymentMethod` to explicitly state that it's scoped to a merchant's vault and not to a customer ID.

crookedneighbor: I think we do need to update the documentation, it's not particularly clear here: https://developers.braintreepayments.com/reference/request/client-token/generate/node#options.fail_on_duplicate_payment_method

src/views/payment-sheet-views/card-view.js
```diff
@@ -396,7 +396,11 @@ CardView.prototype.tokenize = function () {
     });
   }).catch(function (err) {
     self._isTokenizing = false;
-    self.model.reportError(err);
+    if (err.code === 'HOSTED_FIELDS_TOKENIZATION_FAIL_ON_DUPLICATE') {
+      self.model.reportError('hostedFieldsDuplicateCardError');
```
crookedneighbor: If you change the translation key to be `hostedFieldsTokenizationFailOnDuplicateError`, this will work automagically without this if block.

See https://github.com/braintree/braintree-web-drop-in/blob/63cef1448628aa46c9dc5d916b83a7f45f00b93f/src/views/main-view.js#L255-L256

src/translations/en_US.js
```diff
@@ -18,6 +18,7 @@ module.exports = {
   fieldInvalidForPostalCode: 'This postal code is not valid.',
   fieldTooLongForCardholderName: 'Cardholder name must be less than 256 characters.',
   genericError: 'Something went wrong on our end.',
+  hostedFieldsTokenizationFailOnDuplicateError: 'This credit card already exists as a saved payment method.',
```
crookedneighbor: Are we happy with this message?

bluk: :+1:

PR #387, opened 2018-04-11
Fix issue where drop-in would fail to load if ext script is blocked

### LINKED ISSUE #379: GooglePay in not working in 1.10 version
### General information

* SDK/Library version: 1.10
* Environment:  Sandbox
* Browser and OS: Chrome on Windows and Android

### Issue description

When googlePay  option is included in braintree.dropin.create() parameters, like 
googlePay: {
            // merchantId: null, // not needed for sandbox
            totalPriceStatus: 'FINAL',
            totalPrice: payAmount,
            currencyCode: currency
        },
the drop-in Chrome on Windows fails to load and spinner spins forever due to JavaScript load errors:

https://payments.developers.google.com/js/apis/pay.js net::ERR_BLOCKED_BY_CLIENT

At least some error handling is needed.

On Android Chrome drop-in loads, but fails to work:
screenshot_20180328-164828

crookedneighbor:
closes #379

### Summary

Basically, if an external script (such as the Google Pay wrapper) fails to load, Drop-in was just spinning and spinner. This adds a check and resolves the payment method as not being enabled if there is an error in the check for whether or not it is enabled.

### Checklist

- [x] Added a changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ unreleased
 - Use generic error with console log when a payment method fails to set up
 - Fix issue where Mastercard was displayed as MasterCard
 - Allow card form to not be cleared after succesful tokenization with `card.clearFieldsAfterTokenization`
+- Fix issue where Drop-in would fail to load if something blockd an external script from loading (#379)
 
 1.10.0
 ------
--- a/src/dropin-model.js
+++ b/src/dropin-model.js
@@ -258,6 +258,8 @@ function isPaymentOptionEnabled(paymentOption, options) {
   return SheetView.isEnabled({
     client: options.client,
     merchantConfiguration: options.merchantConfiguration
+  }).catch(function () {
+    return Promise.resolve(false);
   });
 }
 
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ unreleased
 - Allow card form to not be cleared after succesful tokenization with `card.clearFieldsAfterTokenization`
 - Fix atob polyfill
 - Update braintree-web to v3.32.1
+- Fix issue where Drop-in would fail to load if something blockd an external script from loading (#379)
```
skunkworks: `blocked` 😀

src/dropin-model.js
```diff
@@ -258,6 +258,8 @@ function isPaymentOptionEnabled(paymentOption, options) {
   return SheetView.isEnabled({
     client: options.client,
     merchantConfiguration: options.merchantConfiguration
+  }).catch(function () {
+    return Promise.resolve(false);
```
crookedneighbor: Should we `console.log` the error here along with the sheet id? Might be good to expose the exception if it gets thrown.

PR #394, opened 2018-05-04
Fix issue where Drop-in would throw if another google script was used

crookedneighbor:
### Summary

We were only checking on the presence of `global.google` when determining if we should skip loading the script. Needed to check for specific values on it instead.

### Checklist

- [x] Added a changelog entry


jaypatel512:
[Unrelated] Just curious if this problem may happen with `PayPal` as well. 

https://github.com/braintree/braintree-web-drop-in/blob/4c0af37bbecb86063bed740af2ae27f1a2d67dc0/src/views/payment-sheet-views/base-paypal-view.js#L112

PR #400, opened 2018-05-16
Update bt-web to v3.34.0

crookedneighbor:
Note: I expect this to fail in IE9-11, but want to see it fail first before including the fix.

### Checklist

- [x] Added a changelog entry


crookedneighbor:
Didn't fail in IE9-11 because we don't strip out builtins like we do in braintree-web when browserifying.

PR #421, opened 2018-08-15
Drop-in would throw undefined error when updating payment method configuration

Rhymond:
### Summary
Fixed issue where undefined error was thrown when payment method is not visible but is set in configuration.

### Checklist

- [x] Added a changelog entry

### DIFF
--- a/src/dropin.js
+++ b/src/dropin.js
@@ -371,7 +371,7 @@ Dropin.prototype._initialize = function (callback) {
  * dropinInstance.updateConfiguration('paypal', 'amount', '10.00');
  */
 Dropin.prototype.updateConfiguration = function (property, key, value) {
-  var view = this._mainView.getView(property);
+  var view;
 
   if (UPDATABLE_CONFIGURATION_OPTIONS.indexOf(property) === -1) {
     return;
@@ -384,6 +384,8 @@ Dropin.prototype.updateConfiguration = function (property, key, value) {
 
     return;
   }
+
+  view = this._mainView.getView(property);
 
   if (!view) {
     return;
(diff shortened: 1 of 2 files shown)


src/dropin.js
```diff
@@ -383,6 +383,10 @@ Dropin.prototype.updateConfiguration = function (property, key, value) {
     return;
   }
 
+  if (this._mainView.getView(property) === undefined) {
```
crookedneighbor: I think this should be:

```js
var view = this._mainView.getView(property);

if (!view) {
  return;
}

view.updateConfiguration(key, value);
```

Rhymond:
This situation occurs at the moment for me then we're trying to update apple pay configuration after price change. If user don't use device which is capable to work with Apple pay, we get `undefined` error. 

I would agree with more descreptive error message, but in any case it's not an error more like warning by my case.

Let me know what do you think it would be correct behaviour and I will update code respectively on your decision.

src/dropin.js
crookedneighbor: This should still be updated.

Rhymond:
Thanks for an answer. 

Yup, that's right. I believe PayPal will work on almost all cases, but we have an issue especially with Apple Pay. 

I'll fix current PR and will push changes for a review.

crookedneighbor:
This needs a test update to cover this case, but we can go ahead and do that once it's merged in if you're not comfortable updating the unit tests.

Rhymond:
@crookedneighbor I will try to update unit tests over the weekend

crookedneighbor:
@Rhymond How's it coming? Do you need us to handle the tests?

Rhymond:
@crookedneighbor sorry, had no time over the weekend. If you can would be great, if you can wait for one more week I can try to do it. 
Thanks.

crookedneighbor:
Merged in with https://github.com/braintree/braintree-web-drop-in/commit/8369e45fba7a003299ce5c2038795da93cd6c417

PR #424, opened 2018-08-31
Fix issue the keyboard could get stuck when entering card details

### LINKED ISSUE #419: iOS Keyboard reopens after loosing focus and closing keyboard
### General information

* SDK/Library version: 1.12.0
* Environment: All
* Browser and iOS: Safari, Chrome 68.0.3440.83 (64-bit) on iOS 11.4.1

### Steps to reproduce

1. Click in the credit card number input field
2. Fill in some numbers(Doesn't have to be valid or the complete number)
3. Click outside the input box (Somewhere on the page, even if there is no elements there)
4. Close keyboard with the Done or the X button
5. Click somewhere and the keyboard reopens

### Issue description
Keyboard continues to pop open when I click anywhere on the page, even after I call `teardown` on the drop UI. If I click on the field again and maintain focus on the element and click `Done` or  `X` the issue stops. Seems to be happening with all fields.

This issue seems to be nearly identical to one with [Strips](https://github.com/stripe/react-stripe-elements/issues/174)

crookedneighbor:
closes #419

### Summary

There's a bug that's happening in web drop-in on iOS devices, where if you click on the field, then click out of it, the keyboard persists.

Click the done link and it goes away, but click on any element on the page and it re-appears.

I figured out what was causing it

```css
input {
  border: none;
}
```

But... why is this affecting how the inputs _inside an iframe_ behave?

Best guess here is that it's a combo bug in iOS's safari, where the css is causing the keyboard not to leave on blur for all things on the page _and_ a bug in ios iframe inputs where they don't give up their focus when clicking on an element outside of the frame.

Combine them and it repeatedly brings up the keyboard.

The solution was to specify a class on the input:

```css
input.braintree-form__raw-input {
  // styles
}
```
 
Don't know why that would matter, since it was nested under a ton of classes already:

```css
.braintree-sheet__content--form .braintree-form__field-group .braintree-form__field .braintree-form__hosted-field input {
  // styles
}
```

### Checklist

- [x] Added a changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ unreleased
 ----------
 - Provide browserified version of Drop-in on npm at `dist/browser/dropin.js`
 - Fix issue where Drop-in would throw an error when updating not presented payment method
-- Fix issue the keyboard could get stuck when entering card details (#419)
+- Fix issue where the keyboard could get stuck when entering card details on iOS (#419)
 
 1.12.0
 ------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ unreleased
 ----------
 - Provide browserified version of Drop-in on npm at `dist/browser/dropin.js`
 - Fix issue where Drop-in would throw an error when updating not presented payment method
+- Fix issue the keyboard could get stuck when entering card details (#419)
 
 1.12.0
 ------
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ unreleased
 ----------
 - Provide browserified version of Drop-in on npm at `dist/browser/dropin.js`
 - Fix issue where Drop-in would throw an error when updating not presented payment method
+- Fix issue the keyboard could get stuck when entering card details (#419)
```
quinnjn: This changelog could use a rewording. 

> Fix issue where the keyboard could get stuck when entering card details (#419)

PR #450, opened 2018-11-03
Patch for Greek translation

StavrosD:
### Checklist

- [1 ] Added file for Greek translation src/translations/el_GR.js


crookedneighbor:
Closing. See https://github.com/braintree/braintree-web-drop-in/pull/300#issuecomment-339451853

> Because Greek is not currently supported in our PayPal flows, we've opted to not support this as an official translation but you can supply your own translations when creating Drop-in. See our [reference documentation](https://braintree.github.io/braintree-web-drop-in/docs/current/module-braintree-web-drop-in.html#create-examples) for an example of passing a translations object into your create call.

crookedneighbor:

PR #465, opened 2019-01-15
Provide error message in PayPal button view

### LINKED ISSUE #433: Feature Request: Indicating that users need to click the inner "PayPal Checkout" button for PayPal and PayPal Credit
### General information

* SDK/Library version: braintree-web-drop-in@1.11.0
* Browser and OS: All

### Issue description

As an integrator of the Braintree Web Drop-in, I want my users to understand that they need to click the "PayPal Checkout" button _inside_ the "PayPal" option so they can continue, without confusion, checking out.

---

This is something that has been brought to my attention recently, and although it's intuitive to me, it seems like it may not be intuitive for all.

In this section:

image

Users are clicking the "PayPal" or "PayPal Credit" options and not realizing they still need to click the "inner button" to complete the PayPal checkout flow:

image

Although I could badger the user with some errors when they try to continue, I was thinking something like the credit card validation errors might be nice: (these are shown when calling `requestPaymentMethod` when no payment method is requestable)

image

I'm imagining something like this but possibly with a different error message: (I'm not very good with those things...)

image

I'm also open to alternative solutions :)

crookedneighbor:
closes #433

### Summary
Currently there is no message to the customer if they press the button that triggers request payment method. This changes it so the view displays a message to the customer about using the PayPal button.

### Concerns

Is the wording acceptable for this? If so, I can request translations for it.

### Checklist

- [x] Added a changelog entry


crookedneighbor:
screen shot 2019-01-16 at 3 03 38 pm

PR #473, opened 2019-02-01
Stop halting Drop-in setup when Data Collector fails to load

crookedneighbor:
### Checklist

- [x] Added a changelog entry


src/lib/data-collector.js
```diff
@@ -28,14 +28,30 @@ DataCollector.prototype.initialize = function () {
     return global.braintree.dataCollector.create(self._config);
   }).then(function (instance) {
     self._instance = instance;
+  }).catch(function (err) {
+    // log the Data Collector setup error
+    // but do not prevent Drop-in from loading
+    self.log(err);
```
jackellenberger: Do we want analytic events around this?

crookedneighbor: Good idea.

crookedneighbor: Added the event. @jackellenberger and @hollabaq86, are you happy with `data-collector.setup-failed`?

PR #481, opened 2019-02-22
Add card vault opt out option

crookedneighbor:
### Summary
Adds ability to opt out of credit card vaulting

### Checklist

- [x] Added a changelog entry


crookedneighbor:
## gif

save card option


## Needs

- [ ] Design review
- [ ] Translation strings (pending design review)

## Concerns

### Confusing UI when not vaulting in a customer with vaulted payment methods

There is an edge case here where if a customer already has saved payment methods, then adds a card and chooses not to vault it, it'll take them to a view where the saved payment methods are shown and the new, not vaulted card will be visible. Reloading Drop-in will remove the card, but it's a bit of a confusing UI. 

I think the real solve is to re-visit the view flow when going to a terminal state (such as a payment method that wasn't vaulted or cannot be vaulted), rather than going back to the payment options view. I think this would be a major version change though.

### Specific to cards

This feature is specific to cards. The only other payment method where it may be applicable is PayPal (since we don't vault venmo on the client and all other payment methods cannot be vaulted on the client), but to accomplish that, we'd have to dynamically change the flow property in the payment resource creation, which is tricky because the checkout flow has different required parameters than the vault flow.

Not impossible to add functionality to the PayPal views, but it'd be considerably trickier.

src/index.js
```diff
@@ -122,6 +122,9 @@ var VERSION = process.env.npm_package_version;
...
  * @param {object} [overrides.styles] The Hosted Fields [`styles` options](http://braintree.github.io/braintree-web/{@pkg bt-web-version}/module-braintree-web_hosted-fields.html#~styleOptions). These can be used to add custom styles to the Hosted Fields iframes. To style the rest of Drop-in, [review the documentation for customizing Drop-in](https://developers.braintreepayments.com/guides/drop-in/customization/javascript/v3#customize-your-ui).
  * @param {boolean} [clearFieldsAfterTokenization=true] When false, the card form will not clear the card data when the customer returns to the card view after a succesful tokenization.
+ * @param {object} [vault] Configuration for vaulting credit cards. Only applies  when using a [client token with a customer id](https://developers.braintreepayments.com/reference/request/client-token/generate/#customer_id).
```
quinnjn: Double space in the description: `applies  when`

src/index.js
```diff
@@ -122,6 +122,9 @@ var VERSION = process.env.npm_package_version;
...
+ * @param {boolean} [vault.showSaveCardToggle=false] When true, the card form will include an option to let the customer decide not to vault the credit card they enter.
+ * @param {boolean} [vault.defaultValueForVaulting=true] Whether or not to have the save card input pre-checked. When set to `false` with `showSaveCardToggle` set to `false`, then cards will not be vaulted upon tokenization.
```
quinnjn: I'm not a big fan of this key name. I'd like the two properties to be pretty closely named to make it obvious. What about:
```
vault.saveCardToggleVisible=false
vault.saveCardToggleDefaultValue=true
```
or
```
vault.saveCardInputVisible
vault.saveCardInputDefaultValue
```

test/unit/views/payment-sheet-views/card-view.js
```diff
@@ -2115,6 +2197,15 @@ describe('CardView', function () {
       });
     });
 
+    it('odes not include `vaulted: true` in tokenization payload if save card input is not checked', function () {
```
quinnjn: s/odes/does/

src/index.js
crookedneighbor: The one trouble with this is that the default value option applies _even_ if the card input is not visible. This allows merchants to choose to display saved payment methods, but not vault cards automatically on the client.

If we change the name to more closely align, with that, we should remove that behavior. Are we :+1: with that?

hollabaq86: I think we should include the option to automatically vault a card on the client. This will help merchants with more complex integrations (verifications and Advanced Fraud Tools)

crookedneighbor: Sorry, by "remove that behavior", I meant, we shouldn't respect this value setting if the option to _show_ the toggle is not enabled.

hollabaq86: Blade and I IRL'd. We think we should keep the current names since the behaviors are related but independent of each other. @quinnjn let us know if that works for you?

quinnjn: Yep that works for me! Resolving

bhushan90:
@crookedneighbor could you please tell me how to add save credit card button in drop-in UI?

hollabaq86:
Hi @bhusan90 please contact Braintree Support for assistance https://help.braintreepayments.com/

bhushan90:
@hollabaq86 I contact them I got a response but now about how to implement this functionality.
it would be good if you can help me to do that

bhushan90:
@hollabaq86  I found the solution and solved the opt-in and opt-out issue.
It was mentioned in Braintree docs
Thank you for support.

rossiter10:
Has there been any update on that edge case mentioned above ("Confusing UI when not vaulting in a customer with vaulted payment methods")?

It seems to be an issue even when there are no existing payment methods. I created a new issue for it here: #598

PR #484, opened 2019-02-25
Add ability to expand "other ways to pay" menu

crookedneighbor:
### Summary

Provides an option to auto-expand other ways to pay menu when using a client token with a customer id. This allows the full range of payment options that are available to be used in Drop-in to be shown, instead of _just_ the saved ones.

### Checklist

- [x] Added a changelog entry


crookedneighbor:
expand

src/dropin-model.js
```diff
@@ -137,6 +137,10 @@ DropinModel.prototype.confirmPaymentMethodDeletion = function (paymentMethod) {
   this._emit('confirmPaymentMethodDeletion', paymentMethod);
 };
 
+DropinModel.prototype.shouldExpandPaymentOptions = function () {
+  return Boolean(this.merchantConfiguration.expandOtherWaysToPay && this.getPaymentMethods().length > 0);
```
jackellenberger: will there ever be a situation where only cards are supported, so `this.getPaymentMethods.length = 1` and we _don't_ want to expand other ways to pay? i.e. should this be kicked up to `> 1`?

crookedneighbor: Good call. This bit is fine as is, but a check for `this.supportedPaymentOptions.length > 1` does need to be added.

vaultwiki:
dropin.js _setUpDependenciesAndViews needs to create this._mainView before the setup of dataCollector and 3dsecure, or asyncDependenciesReady can fire before this._mainView exists, which causes the expandPaymentOptions call there to fail against an undefined object.

This reveals a deeper problem that the individual setups of dataCollector and 3dsecure may finish before the other one queues, which can lead to asyncDependenciesReady firing while one is about to enter the queue. These should be changed so that the initialization is returned as a callback from the individual _setup functions and handled by a finalization step of _setUpDependenciesAndViews. This could explain the random failures of dropin to initialize that I have experienced in the past.

crookedneighbor:
📕 (I trust y'all to handle this when you get to it) 📕

PR #502, opened 2019-05-09
Remove sass

gesa:
### Summary

node-sass has the kind of vulnerability that… may take a moment to resolve:  

Seems as good a time as any to swap out Sass for the more javascript-y less anyway.

### Checklist

- [x] Added a changelog entry


src/less/three-d-secure.less
```diff
@@ -35,8 +35,6 @@
     border-radius: 3px;
     margin: auto;
     height: 450px;
-    width: 400px;
```
hollabaq86: why are these getting removed?

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 CHANGELOG
 =========
 
+Unreleased
+------
+- Replace css preprocessor, from sass to less
```
hollabaq86: Can we add links to `sass` and `less` repo pages?

src/less/three-d-secure.less
crookedneighbor: I'm curious about that too.

CHANGELOG.md
crookedneighbor: Ditto

src/less/three-d-secure.less
gesa: See lines 33-34 of this file (duplicates)

crookedneighbor: :+1:

hollabaq86: Ahhhhhhh thanks, GH cut that part off

gesa:
The only option I considered outside of Less was vanilla CSS, but CSS variables have a long way to go, and we utilize variables quite a bit. Also nesting/un-nesting is a task fraught with potential human error, where Sass -> Less conversion is pretty easy. On that note, I diffed built css between the two preprocessors, there were no differences outside of whitespace.

PR #505, opened 2019-05-17
Replace placeholders with strings in dist

gesa:
### Summary

The JSDoc plugin was replacing placeholders in the documentation, but
ultimately that left placeholders all over the dist code outside of the
docs. This change interpolates the current version numbers not just in
the docs, but also all of the comments that the docs are generated
from.

Also apparently I accidentally wrote "statistics" instead of "statics"
in one place on the gulp@4 upgrade.

### Checklist

- [x] Added a changelog entry


gulpfile.js
```diff
@@ -180,16 +181,20 @@ function build() {
...
 }
 
-function _replaceVersionInFile(filename) {
-  var updatedFile = fs.readFileSync(filename, 'utf-8').replace(/@VERSION/g, VERSION);
-
-  fs.writeFileSync(filename, updatedFile, 'utf-8');
+function cleanUpVersions() {
+  return src('dist/**/*.{js,html}')
+    .pipe(replace('{@pkg version}', VERSION))
+    .pipe(replace('{@pkg bt-web-version}', BT_WEB_VERSION))
```
crookedneighbor: If we're now doing this here, should we remove this plugin? https://github.com/braintree/braintree-web-drop-in/blob/c8bc81b64def5ef739ea5b42db0c65c7d1fc75ce/jsdoc/version-interpolator-plugin.js

gesa: No, because when we do, jsdoc no longer takes markdown links and turns them into anchor tags

PR #509, opened 2019-05-24
Update sauce and capybara

crookedneighbor:
### Summary

* Updates Sauce labs to the latest binary.
* Updates capybara to v3

### Checklist


.ruby-version
```diff
@@ -0,0 +1 @@
+ruby-2.4.3
```
quinnjn: I'm 👍 with this, but I also saw you can define it in the Gemfile which is cool: https://rvm.io/workflow/projects#supported-files

.travis.yml
```diff
@@ -1,12 +1,9 @@
 language: node_js
 before_install:
-# See https://github.com/travis-ci/travis-ci/issues/5092#issuecomment-171777646
-# Dirty hack for https://github.com/travis-ci/travis-ci/issues/5092
-- |
-  PATH=`echo $PATH | sed "s/\.\/node_modules\/\.bin//g"`; export PATH=$PATH:./node_modules/.bin; touch .env; if [ $IS_INTEGRATION_TEST ]; then rvm install ruby-2.3.0 && rvm use ruby-2.3.0 && bundle install; else echo "Not an integration test. Skipping Ruby setup"; fi
+  touch .env;
+  if [ $IS_INTEGRATION_TEST ]; then rvm install ruby-2.4.3 --binary && rvm use ruby-2.4.3 && bundle install; else echo "Not an integration test. Skipping Ruby setup"; fi
```
quinnjn: rvm should be able to pick up `2.4.3` from our .ruby-verison file. We could switch this to a `before_script` task to take advantage of that file.

Gemfile.lock
```diff
@@ -3,103 +3,84 @@ GEM
   specs:
     addressable (2.6.0)
       public_suffix (>= 2.0.2, < 4.0)
-    capybara (2.7.1)
```
quinnjn: Oof we need a `Gemfile.lock -diff` line in our `.gitattributes` to not show this lock file for PRs.

.ruby-version
crookedneighbor: Just tried this and got:

```
RVM used your Gemfile for selecting Ruby, it is all fine - Heroku does that too,
you can ignore these warnings with 'rvm rvmrc warning ignore ./Gemfile'.
To ignore the warning for all files run 'rvm rvmrc warning ignore allGemfiles'.

Using /home/admin/.rvm/gems/ruby-2.4.3
```

Which means I think we'd need to update the rvm config on the cpair image if we wanted to use the Gemfile instead.

.travis.yml
crookedneighbor: I did this, but it failed to find the `.ruby-version` file. Wonder if this gets run on a different path than the repo itself.

Gemfile.lock
crookedneighbor: Done.

crookedneighbor: Though, it looks like github doesn't respect that. Local git did.

PR #521, opened 2019-07-02
hint audit fixes

gesa:
### Summary

Fixes inspired by hint inspections. There are still a whole bunch of css-related warnings that beg pairing time, as well as some server-side warnings that likely should be eliminated altogether for this repo but may be worth a quick look before doing so.

Would [this typo fix](521/files#diff-1425077fec6915abda9292c4827dc1e6R17) be considered a breaking change?
### Checklist

- ~Added a changelog entry~


crookedneighbor:
Re: the typo, I don't think so. I'm curious though about the styling, do we have specific styling for that class? (can look myself later)

PR #523, opened 2019-07-10
3ds 2

crookedneighbor:
### Summary

Adds cursory 3DS 2 support to Drop-in.

3DS 2 fields are passed in as an object to `requestPaymentMethod`.

In the future:

* Add ability to collect info via a form _within_ Drop-in
* Add ability to pass in a custom `onLookupComplete` function so 3DS verification can be cancelled if lookup info does not match required parameters

### Checklist

- [x] Added a changelog entry


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Unreleased
 - Update braintree-web to v3.46.0
 - Prevent Drop-in from firing requestable events prematurely (#511)
 - Add `off` method for unsubscribing from events without tearing down
+- Add ability to use 3DS 2 flow
```
hollabaq86:
```suggestion
- Add 3DS version 2 support
```

src/dropin.js
```diff
@@ -619,7 +619,12 @@ Dropin.prototype._handleAppSwitch = function () {
...
+ * @param {object} [options] All options for requesting a payment method.
+ * @param {object} [options.threeDSecure] Any of the options in the [Braintree 3D Secure client reference](https://braintree.github.io/braintree-web/{@pkg bt-web-version}/ThreeDSecure.html#verifyCard) except for `nonce`, `bin`, and `onLookupComplete`. If `amount` is provided, it will override the value of `amount` in the [3D Secure create options](module-braintree-web-drop-in.html#~threeDSecureOptions). The more options provided, the more likely the customer will not need answer a 3DS challenge. The recomended fields for achieving a 3DS 2 lookup are:
```
hollabaq86:
```suggestion
 * @param {object} [options.threeDSecure] Any of the options in the [Braintree 3D Secure client reference](https://braintree.github.io/braintree-web/{@pkg bt-web-version}/ThreeDSecure.html#verifyCard) except for `nonce`, `bin`, and `onLookupComplete`. If `amount` is provided, it will override the value of `amount` in the [3D Secure create options](module-braintree-web-drop-in.html#~threeDSecureOptions). The more options provided, the more likely the customer will not need to answer a 3DS challenge. The recommended fields for achieving a 3DS v2 verification are:
```

src/lib/three-d-secure.js
```diff
@@ -1,152 +1,47 @@
...
-
-    if (err.type === 'THREE_D_SECURE_CANCELLED') {
-      return Promise.resolve(err.payload);
+ThreeDSecure.prototype.verify = function (payload, merchantProvidedData) {
+  var verifyOptions = assign({
+    amount: this._config.amount
+  }, merchantProvidedData, {
+    nonce: payload.nonce,
+    bin: payload.details.bin,
+    // TODO in the future, we will allow
```
hollabaq86: do we want to keep this TODO note in?

crookedneighbor: I think so, since it's something we want to add imminently, I'd like it to keep showing up as a warning in eslint.

src/dropin.js
```diff
@@ -619,7 +619,12 @@ Dropin.prototype._handleAppSwitch = function () {
...
+ * @param {object} [options] All options for requesting a payment method.
+ * @param {object} [options.threeDSecure] Any of the options in the [Braintree 3D Secure client reference](https://braintree.github.io/braintree-web/{@pkg bt-web-version}/ThreeDSecure.html#verifyCard) except for `nonce`, `bin`, and `onLookupComplete`. If `amount` is provided, it will override the value of `amount` in the [3D Secure create options](module-braintree-web-drop-in.html#~threeDSecureOptions). The more options provided, the more likely the customer will not need to answer a 3DS challenge. The recommended fields for achieving a 3DS v2 verification are:
```
demerino: Do we want to be this specific with comments like? I'd feel more comfortable pointing towards a `best practices` guide that would be evolving (maybe the adoption guide).
> The more options provided, the more likely the customer will not need to answer a 3DS challenge. The recommended fields for achieving a 3DS v2 verification are

Not a big deal tho since these are currently pretty accurate statements.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Unreleased
 - Prevent Drop-in from firing requestable events prematurely (#511)
 - Add `off` method for unsubscribing from events without tearing down
 - Clean up extraneous css rules
+- Add 3DS version 2 support
```
demerino: In our docs we now officially call it `3DS 2`: https://developers.braintreepayments.com/guides/3d-secure/migration/android/v3

Not to bike shed this more tho.

src/dropin.js
crookedneighbor: I think that's a good idea, but let's wait to do it until the docs app is a little more stable. It's in a bit of flux right now as we transition from the legacy guide to the 3DS 2 guide.

Issue #546, opened 2019-08-23
Needless scroll bars for 3DS challenge


liamsharp:
### General information

* SDK/Library version: 1.20.0
* Environment: Sandbox
* Browser and OS: Safari Version 12.1.1 (14607.2.6.1.1) on macOS 10.14.5

### Issue description

Using the CodePen [here](https://codepen.io/braintree/pen/KjWqGx).

Version 1.19.0 displays a 3DS challenge like this:

<img width="1721" alt="codepen_19" src="https://user-images.githubusercontent.com/6429288/63593412-db401400-c5ab-11e9-8d2d-12faf0f02993.png">

However version 1.20.0 shows it like this:

<img width="1721" alt="codepen_20" src="https://user-images.githubusercontent.com/6429288/63593424-e5621280-c5ab-11e9-8c93-aa1e4d26bd60.png">

Is this an intentional change?

It's obvious you have to scroll in 1.19.0, but in 1.20.0 it's unexpected and needless.

You can switch between 1.19.0 and 1.20.0 in the CodePen by going to Settings, JavaScript, switching the version, clicking close then the Run button on the main UI.

<img width="527" alt="codepen_change_version" src="https://user-images.githubusercontent.com/6429288/63593791-a97b7d00-c5ac-11e9-9c6f-44f49d1db1bb.png">

jamesefl:
I much prefer the appearance of Version 1.19.0 to Version 1.20.0.

Without the semi-transparent background it doesn't look good on a computer monitor and it looks much better having the scroll bar near the content.

wagonli:
Same issue here on Linux Fedora 29 with firefox and chrome.
This regression was introduced by commit 99d74fc4b6c3f1fe0d130c96945a43b69e4e5274, the style of the iframe content is set to use all the screen space to be displayed.

crookedneighbor:
This will be fixed in the next release. Gonna re-open this for now and close it again when the release is out.

crookedneighbor:
This is fixed in 1.20.1

PR #548, opened 2019-08-27
Fix Cardinal iframe style

wagonli:
### Summary

Fix 3DSecure popup sizing issue introduced in 1.20.0
This should fix part of #546.

### Checklist

- Remove top, left, bottom and right iframe container prositioning along with translation transform

### DIFF
--- a/src/less/cardinal-overrides.less
+++ b/src/less/cardinal-overrides.less
@@ -5,15 +5,6 @@
 // to fix these issues, we can remove these styles
 // fix found here: https://davidwalsh.name/scroll-iframes-ios
 #Cardinal-Modal {
-	top: 10px !important;
-	bottom: 10px !important;
-	left: 10px !important;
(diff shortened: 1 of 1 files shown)


crookedneighbor:
Thanks for the PR.

This was an intentional change to override the styles to overcome an iOS scrolling bug. We'll review your changes (to make sure the workaround still works) and add the full screen view as an option a merchant can opt into instead.

wagonli:
@crookedneighbor Ok, we thought that the fullscreen popup was an unintentional side effect of the iOS fix as it was not explained in the dropin changelog.
This new style is fit for mobile but is not really well suited for desktop as the banks 3DSecure pages are made for a small viewport.

crookedneighbor:
@wagonli Here's a PR to address the issue: https://github.com/braintree/braintree-web-drop-in/pull/554

PR #555, opened 2019-09-06
Fix issue where css class name could not be passed in as a HF override

### LINKED ISSUE #535: Passing CSS class name for override throws error
### General information

* SDK/Library version: 1.19.0
* Environment: sandbox
* Browser and OS Chrome 75 / Linux

### Issue description

According to the hosted fields [documentation](https://braintree.github.io/braintree-web/current/module-braintree-web_hosted-fields.html#~styleOptions) (which this plugin refers to for styling help) I should be able to pass a CSS class name for a style override, such as:

```javascript
dropin.create({
      authorization: context.braintreeAuthorizationKey,  
      container: '#braintree-dropin-container',  
      card: {  
        overrides: {  
          styles: {  
            'number': 'TextInput',  
          },  
          fields: {  
            number: {maskInput: {showLastFour: true}},  
            cvv: {maskInput: true},  
          },  
        },  
      },  
    }, (createErr, instance) => { });  
```

However when I try this I receive the following exception:
`Cannot assign to read only property '0' of string 'TextInput'
    at https://localhost:3000/webpack/app.js:35704:32
    at Array.forEach (<anonymous>)
    at normalizeStyles (https://localhost:3000/webpack/app.js:35701:23)
    at https://localhost:3000/webpack/app.js:35285:7
    at Array.forEach (<anonymous>)
    at CardView._generateHostedFieldsOptions (https://localhost:3000/webpack/app.js:35278:35)
    at CardView.initialize (https://localhost:3000/webpack/app.js:35069:20)`

Is this functionality intended to be supported?

Am I misunderstanding the hosted field documentation?

crookedneighbor:
closes #535

### Summary

Hosted Fields supports passing in a class name as a style. (under the hood we compute the style from the class name). Drop-in didn't assume that and would break if a string was passed instead of a style object.

This PR fixes that.

### Checklist

- [x] Added a changelog entry

### DIFF
--- a/test/unit/views/payment-sheet-views/card-view.js
+++ b/test/unit/views/payment-sheet-views/card-view.js
@@ -660,6 +660,37 @@ describe('CardView', function () {
         });
       });
     });
+
+    it('allows overriding styles options with class name for hosted fields', function () {
+      var hostedFieldsConfiguredStyles;
+
+      this.model.merchantConfiguration.card = {
+        overrides: {
+          styles: {
+            input: 'class-name',
+            ':focus': 'focus-class'
+          }
+        }
+      };
+
+      this.view = new CardView({
+        element: this.element,
+        mainView: this.mainView,
+        model: this.model,
+        client: this.client,
+        strings: strings
+      });
+
+      return this.view.initialize().then(function () {
+        hostedFieldsConfiguredStyles = hostedFields.create.lastCall.args[0].styles;
+
+        expect(hostedFieldsConfiguredStyles.input).to.equal('class-name');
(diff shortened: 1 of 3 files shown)


test/unit/views/payment-sheet-views/card-view.js
```diff
@@ -660,6 +660,37 @@ describe('CardView', function () {
         });
       });
     });
+
+    it('allows overriding styles options with class name for hosted fields', function () {
```
gesa: FUN FACT jsdom is _very bad_ at `getComputedStyle` so we have that to look forward to if we ever change testing loaders

PR #560, opened 2019-09-13
Czech language translation

radimhornicek:
### Checklist

- [ ] Added a changelog entry


crookedneighbor:
The PR template states:

>NOTE: We cannot accept language translation PRs. We support the same languages that are supported by PayPal, and have a dedicated localization team to provide the translations.
>
>If there is an error in a specific translation, you may open an issue and we will escalate it to the localization team.

If you would like to use an unofficial translation, you can pass in your own translations object. See the translations example here: https://braintree.github.io/braintree-web-drop-in/docs/current/module-braintree-web-drop-in.html#.create

crookedneighbor:

PR #578, opened 2019-11-25
Automatically disable future funding sources

crookedneighbor:
### Summary

Prevents non-bt allowed payment buttons from auto-appearing in the PayPal views.

### Checklist

- [x] Added a changelog entry

- @crookedneighbor

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ unreleased
   - Google Pay: Add support for `isNetworkTokenized` param in `parseResponse` method
   - Card Form: Fix issue where pasting a card number over an Amex number could cut off the last digit
   - PayPal: Add support for shipping options (see https://braintree.github.io/braintree-web/current/PayPalCheckout.html#createPayment)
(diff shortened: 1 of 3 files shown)


nampas:
@crookedneighbor Do you know how long this was an issue for? We're on v1.20.2, and I'm 99% sure that we weren't seeing Venmo in the PayPal view when we initially upgraded to that version in October. 

We will be upgrading 1.21.0 (or newer) asap.

crookedneighbor:
@nampas the problem was that the checkout.js team would ramp up and ramp down whether they were showing alt payment buttons. So the behavior was unpredictable. This PR ensures that it will not show alternative modes of payment.

I'd recommend upgrading to 1.22.0, which is the latest.

nampas:
@crookedneighbor Gotcha. We may need some of those ramp up / down times for our analytics team to see if this has had any effect on our conversion, but if that's the case I'll write a support ticket.

Thanks for the quick reply!

PR #587, opened 2020-01-22
Slovak and Czech language added

maros-852:


scannillo:
This note is in our PR template:

We cannot accept language translation PRs. We support the same languages that are supported by PayPal, and have a dedicated localization team to provide the translations. 

If there is an error in a specific translation, you may open an issue and we will escalate it to the localization team. 

You may provide your own custom translations by using the `translations` parameter when creating the Drop-in instance: https://braintree.github.io/braintree-web-drop-in/docs/current/module-braintree-web-drop-in.html#.create

Issue #156, opened 2017-05-10
paymentMethodRequestable Triggering Prematurely


bkonia:
I implemented the new paymentMethodRequestable event on my form. It works perfectly when adding a PayPal account. However, when adding a credit card, if you have the zip code field enabled, the event triggers when the user enters the third digit of his zip code.  This is incorrect since the credit card is not requestable until after the user clicks the Submit button.

In my case, I'm using the event to retrieve the list of payment methods and update a dropdown with the available methods. Since the event triggers before the credit card has been added, the dropdown updates prematurely and the new credit card does not show up until after the page has been refreshed.

crookedneighbor:
For PayPal accounts, you can call `requestPaymentMethod` on the dropin instance only after the PayPal account has been authenticated, otherwise, it'll error. That's why the event fires after authentication completes.

For credit cards, you are able to attempt to call `requestPaymentMethod` when the card passes client side validation. Client side validation for postal code is that it is longer than 3 characters. (some international postal codes can have as few as 3 characters) 

If you need special logic for credit cards, you can leverage the data passed back with the event object:

```js
dropinInstance.on('paymentMethodRequestable', function (event) {
  if (event.type === 'CreditCard') {
    // display submit button so payment method can be requested by the user once postal code is done being entered
  } else {
    // whatever logic you are currently doing
  }
```

We'll look into adding a configurable `minLength` property to postal code in [Hosted Fields](http://braintree.github.io/braintree-web/current/HostedFields.html) (what Drop-in is using to create the card form) so that the client side validation will not report that it is valid until that `minLength` value is reached.

PR #589, opened 2020-01-28
clarify paymentMethodRequestable docs

hollabaq86:
### Summary

Braintree Support gets regular inbounds where merchants are calling `requestPaymentMethod` as soon as the `paymentMethodRequestable` event is received, which can result in tokenizations being attempted before the user has completed entering a US postal code(US postal codes are 5 digits, but we have to support intl postal codes that have fewer numbers. See #156).

This PR attempts to clarify the `paymentMethodRequestable` event in the reference docs. We can also document this behavior more thoroughly in our developer docs 

- @hollabaq86


src/dropin.js
```diff
@@ -219,7 +219,7 @@ HAS_RAW_PAYMENT_DATA[constants.paymentMethodTypes.applePay] = true;
  */
 
 /**
- * This event is emitted when the payment method available in Drop-in changes. This includes when the state of Drop-in transitions from having no payment method available to having a payment method available and when the payment method available changes. This event is not fired if there is no payment method available on initialization. To check if there is a payment method requestable on initialization, use {@link Dropin#isPaymentMethodRequestable|`isPaymentMethodRequestable`}.
+ * This event is emitted when the payment method available in Drop-in changes. This includes when the state of Drop-in transitions from having no payment method available to having a payment method available and when the payment method available changes. This event is not fired if there is no payment method available on initialization. Note: auto-submit behavior should not be dependent on this event. To check if there is a payment method requestable on initialization, use {@link Dropin#isPaymentMethodRequestable|`isPaymentMethodRequestable`}.
```
crookedneighbor: Auto submit behavior is the exact case for use here, especially for non-cc payment methods. CC payment methods can do it too, just need to be careful if the fraud rules are set up to include postal code, since it'll fire too early.

The code example specifically shows how to check if it's a payment method that should be allowed to submit automatically.

PR #602, opened 2020-03-13
"test" to "it", also some subtle cleanup (mostly whitespace)

gesa:
### Summary

also i rearranged some imports

### Checklist

- ~Added a changelog entry~

- @gesa

### DIFF
--- a/test/.eslintrc
+++ b/test/.eslintrc
@@ -10,7 +10,6 @@
   },
   "globals":{
     "global": true,
-    "jest": true,
     "$": true,
     "$$": true,
     "browser": true
@@ -24,6 +23,7 @@
         "arraysInObjects": false,
         "objectsInObjects": false
       }
-    ]
+    ],
+    "no-console": 0
   }
 }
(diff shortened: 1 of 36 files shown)


test/.eslintrc
```diff
@@ -24,6 +23,7 @@
         "arraysInObjects": false,
         "objectsInObjects": false
       }
-    ]
+    ],
+    "no-console": 0
```
gesa: should i maybe move this to only apply to the file dropin-model?

test/unit/views/base-view.js
```diff
@@ -13,31 +12,24 @@ describe('BaseView', () => {
...
 
-  describe('teardown', () => {
-    test('returns a resolved promise', () => {
-      const view = new BaseView();
-      const promise = view.teardown();
-
-      expect(promise).toBeInstanceOf(Promise);
-    });
-  });
+  describe.todo('teardown');
```
crookedneighbor: getting `describe.todo is not a function` in the tests

crookedneighbor: Pushed up a fix for this.

PR #627, opened 2020-06-02
Add 3DS support for non-network tokenized Google Pay cards

demerino:
### Summary
Runs 3DS verification for non-network tokenized Google Pay cards if 3DS is enabled as well.

### Checklist

- [x] Added a changelog entry

@billwerges 
@demerino

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ CHANGELOG
 
 unreleased
 ----------
+- Add 3DS support for non-network tokenized Google Pay cards
 - Update braintree-web to v3.62.1
 
 1.22.1
--- a/src/dropin.js
+++ b/src/dropin.js
@@ -97,6 +97,8 @@ HAS_RAW_PAYMENT_DATA[constants.paymentMethodTypes.applePay] = true;
  * @property {string} details.cardType Type of card, ex: Visa, Mastercard.
  * @property {string} details.lastFour The last 4 digits of the card.
  * @property {string} details.lastTwo The last 2 digits of the card.
+ * @property {boolean} details.isNetworkTokenized True if the card is network tokenized.
+ * @property {string} details.bin First six digits of card number.
  * @property {external:GooglePayPaymentData} details.rawPaymentData The raw response back from the Google Pay flow, which includes shipping address, phone and email if passed in as required parameters.
  * @property {string} type The payment method type, always `AndroidPayCard` when the method requested is a Google Pay Card.
  * @property {object} binData Information about the card based on the bin. Documented {@link Dropin~binData|here}.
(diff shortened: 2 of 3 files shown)


test/unit/dropin.js
```diff
@@ -1194,7 +1194,7 @@ describe('Dropin', () => {
     });
 
     test(
-      'does not call 3D Secure if payment method is not a credit card',
+      'does not call 3D Secure if payment method is not a credit card or google pay',
```
scannillo, line 1197:
```suggestion
      'does not call 3D Secure if payment method is not a credit card or network tokenized google pay',
```

scannillo: take or leave this comment

src/dropin.js
```diff
@@ -97,6 +97,8 @@ HAS_RAW_PAYMENT_DATA[constants.paymentMethodTypes.applePay] = true;
  * @property {string} details.cardType Type of card, ex: Visa, Mastercard.
  * @property {string} details.lastFour The last 4 digits of the card.
  * @property {string} details.lastTwo The last 2 digits of the card.
+ * @property {boolean} details.isNetworkTokenized True if the card is network tokenized.
```
hollabaq86, line 100:
```suggestion
 * @property {boolean} details.isNetworkTokenized True if the card is network tokenized. A network tokenized card is a generated virtual card with a device-specific account number (DPAN) that is used in place of the underlying source card.
```

hollabaq86: Pulled this description from our [reference docs](https://developers.braintreepayments.com/reference/response/android-pay-card/java#is_network_tokenized), this is a term that gets frequent inbounds from confused merchants.

src/dropin.js
```diff
@@ -697,7 +699,7 @@ Dropin.prototype.requestPaymentMethod = function (options) {
   options = options || {};
 
   return this._mainView.requestPaymentMethod().then(function (payload) {
-    if (self._threeDSecure && payload.type === constants.paymentMethodTypes.card && payload.liabilityShifted == null) {
+    if (self._shouldPerformThreeDSecureVerification(payload)) {
```
crookedneighbor, line 702: :+1: on pulling this logic out to a helper method.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ CHANGELOG
 
 unreleased
 ----------
+- Add 3DS support for non-network tokenized Google Pay cards
```
crookedneighbor, line 6: Do we think it's fine for 3ds to just start running for some Google Pay cards? Or should it be opt in instead? I'm conflicted about it.

src/dropin.js
crookedneighbor: Should we be even more explicit and say that 3ds will run against these cards automatically (if 3ds is enabeld) if they are not network tokenized?

CHANGELOG.md
crookedneighbor: :+1:

Issue #635, opened 2020-07-08
Not recieving the 3dsecure information in payload


AndresPrietoTucasi:
### General information

* SDK/Library version: <script src="https://js.braintreegateway.com/web/dropin/1.22.1/js/dropin.min.js"></script>
* Environment: Sandbox
* Browser and OS Google Chrome is up to date Version 83.0.4103.116 (Official Build) (64-bit) , windows 10

### Issue description

I recently followed the documentation available here:
https://developers.braintreepayments.com/guides/3d-secure/client-side/javascript/v3
To extend my current drop-in implementation to include 3D secure, from what i could see all i had to add was the "threeDSecure: true" on creation of the form and pass "threeDSecure: threeDSecureParameters" to the "requestPaymentMethod" function

Having done the above the form requesting "Purchase Authentication" appears and when i fill in the one time password provided in sandbox mode my requestPaymentMethod function is hit. The problem im having is that the documentation states i should be able to access payload.threeDSecureInfo.threeDSecureAuthenticationId which i need to send to my server.
What i am finding is that I get the following:
![image](https://user-images.githubusercontent.com/68009089/86910734-0ffc8180-c112-11ea-8632-ba4f62767841.png)
Which as you can see doesn't include the 3d secure information, but what i find weird is that if i look in the network tab in chrome i can see the following:
![image](https://user-images.githubusercontent.com/68009089/86910887-4c2fe200-c112-11ea-9f4f-e96e5637fad3.png)
So the request is being made and the data is being returned, but im currently not able to access it. Am i missing something?

crookedneighbor:
You're right, that's a mistake. We'll see about exposing it for the next release.

AndresPrietoTucasi:
@crookedneighbor Thank you for the update, I just wanted to clarify, does this mean that 3d secure does not work at the moment in the drop in UI? and how often do you make releases?

crookedneighbor:
It does work, it just doesn't support the specific advanced feature you are mentioning. Most merchants don't need the 3ds info on the client side, all they need is the nonce. They can use that nonce to retrieve the additional 3ds info on their server using: https://developers.braintreepayments.com/reference/request/payment-method-nonce/find

Exposing that additional 3ds info on the client is just for convenience, you should always use the payment method nonce find method on your server to validate that the 3ds info you have received has not been modified by the client before sending to your server.

AndresPrietoTucasi:
[...]

PR #636, opened 2020-07-08
Add `threeDSecureInfo` to 3ds payload

crookedneighbor:
### Summary

Closes https://github.com/braintree/braintree-web-drop-in/issues/635

### Checklist

- [x] Added a changelog entry

- @crookedneighbor

### DIFF
--- a/test/unit/dropin.js
+++ b/test/unit/dropin.js
@@ -1400,7 +1400,10 @@ describe('Dropin', () => {
           verify: jest.fn().mockResolvedValue({
             nonce: 'new-nonce',
             liabilityShifted: true,
-            liabilityShiftPossible: true
+            liabilityShiftPossible: true,
+            threeDSecureInfo: {
+              threeDSecureAuthenticationId: 'id'
+            }
           })
         };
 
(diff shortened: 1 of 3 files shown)


test/unit/dropin.js
```diff
@@ -1400,7 +1400,10 @@ describe('Dropin', () => {
           verify: jest.fn().mockResolvedValue({
             nonce: 'new-nonce',
             liabilityShifted: true,
-            liabilityShiftPossible: true
+            liabilityShiftPossible: true,
+            threeDSecureInfo: {
+              threeDSecureAuthenticationId: 'id'
```
scannillo, line 1405: In the reference docs (https://braintree.github.io/braintree-web/3.63.0/ThreeDSecure.html#~verifyPayload), the only properties listed on the `threeDSecureInfo` are `liabilityShiftPossible` and `liabilityShifted`. Why do we have the property `threeDSecureAuthenticationId` here?

crookedneighbor: Good callout!

We mention this property specifically in the developer docs: (see for it in here: https://developers.braintreepayments.com/guides/3d-secure/client-side/javascript/v3#drop-in-ui )

I've asked our fraud tools team what things we should be exposing in addition to those 2 in the client reference docs. We'll hotfix the docs for this version when I get an answer.

Issue #614, opened 2020-04-13
HOSTED_FIELDS_TIMEOUT; braintree inside iframe and shadow DOM


phips28:
### General information

* SDK/Library version: v1.16.1 & v1.22.1
* Environment: Production
* Browser and OS Chrome Version 80.0.3987.163 (Official Build) (64-bit)

### Issue description

We host an iframe with the checkout process inside a shadow DOM.  But it fails with the following Error: `{name: "BraintreeError", code: "HOSTED_FIELDS_TIMEOUT", message: "Hosted Fields timed out when attempting to set up.", type: "UNKNOWN", details: undefined}`

Our DOM structure:

- `html`
  - `body`
  - ...
  - `shadow root`
     - ...
     - iframe (`<iframe src="https://webviews.dock.bot/qe/address?placeId=5e79f33fbc72b5c3ace13294&&source=QE&&rate=45&&itemName=Bug Out Bag" sandbox="allow-forms allow-pointer-lock allow-popups allow-same-origin allow-scripts allow-top-navigation" allowfullscreen="" width="100%" height="100%" data-openedfrombotcopy="true"></iframe>`) - website with react and braintree-web-drop-in
        - `div.braintree-dropin-react`
           - `div.braintree-dropin-react-form`

I have read through the code (braintree-web-drop-in & braintree-web), and found that iframe stuff. But I dont understand why you need to access the parent if the checkout is hosted within the iframe.

When using the iframe without the wrapping shadow DOM everything works. (But I cant change this because we are using another plugin that creates the shadow DOM)

---

Related issue: https://github.com/braintree/braintree-web-drop-in/issues/296

But that workaround doesnt work for us, as we use another plugin that creates the shadow DOM. 

Any other idea how to fix this?

gesa:
Hey @phips28. It's not clear to me what you meant in a few different places. What were you referring to when you mentioned "that iframe stuff"?

There are a lot of variables here that may be compounding issues. We've seen issues in the past with nested iframes that have sandboxing applied. If the drop-in is being initialized inside an iframe that is a fully-formed document, I wouldn't _expect_ the shadow dom to be the issue. Can you provide a simplified example of this problem being triggered?

phips28:
@gesa hello, thx for your response.

Here is a video of the issue https://www.dropbox.com/s/o05r4nw79rag4cd/quarantine%20%281%29.mp4?dl=0

And here the steps to reproduce -> 
[...]

PR #647, opened 2020-08-13
Add support for shadow DOM in card view

crookedneighbor:
addresses #614

### Summary

Only adds support for the card view.

Depends on an unreleased version of braintree-web to work completely.

* PayPal - should have support in the newest SDK, but not the one we're using currently, so this will have to wait til lv2
* Google Pay - mostly works, just need to figure out how to get the styles to load correctly in the shadow dom instead of in the head. May be as simple as adding the google pay script to the shadow dom, but may need to contact google pay team about it
* Apple Pay - unknown
* Venmo - unknown
* 3D Secure - unknown
* Data Collector - unknown

### Checklist

- [x] Added a changelog entry

- @crookedneighbor

### DIFF
--- a/test/unit/dropin-model.js
+++ b/test/unit/dropin-model.js
@@ -29,6 +29,7 @@ describe('DropinModel', () => {
     };
     jest.spyOn(vaultManager, 'create').mockResolvedValue(testContext.vaultManager);
     testContext.modelOptions = {
+      container: document.createElement('div'),
       client: fake.client(testContext.configuration),
       componentID: 'foo123',
       merchantConfiguration: {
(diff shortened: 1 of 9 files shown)


test/unit/dropin-model.js
```diff
@@ -70,6 +71,28 @@ describe('DropinModel', () => {
       expect(model.merchantConfiguration).toBe(testContext.modelOptions.merchantConfiguration);
     });
 
+    test('is sets isInShadowDom to false when the container is not in the shadow DOM', () => {
```
gesa, line 74:
```suggestion
    test('it sets isInShadowDom to false when the container is not in the shadow DOM', () => {
```

test/unit/dropin-model.js
```diff
@@ -70,6 +71,28 @@ describe('DropinModel', () => {
...
+      const model = new DropinModel(testContext.modelOptions);
+
+      expect(model.isInShadowDom).toBe(false);
+    });
+
+    test('is sets isInShadowDom to true when the container is in the shadow DOM', () => {
```
gesa, line 80:
```suggestion
    test('it sets isInShadowDom to true when the container is in the shadow DOM', () => {
```

PR #649, opened 2020-08-13
Add method to return the available payment options

### LINKED ISSUE #594: Add a method to Drop-in UI to return supported payment types
### General information

* SDK/Library version: 1.22.0
* Browser and OS: any browser

### Feature request

I have a requirement to be able to detect whether or not Venmo is supported on the client browser, in order to provide further instructions to the user.  This would be useful for any other payment method as well.

I currently use a workaround where I check an array within "clientInstance" parameter, however this is not guaranteed to work with future releases.

`clientInstance._mainView.paymentMethodsViews.model.supportedPaymentOptions.includes('venmo')`

Please add a method like "clientInstance.isSupportedPaymentMethod(_paymentMethod_)".

See support ticket #2723511 for reference.

crookedneighbor:
closes #594

### Summary

Presents an array of payment options that have been presented to the customer.

### Checklist

- [x] Added a changelog entry

- @crookedneighbor

### DIFF
--- a/src/dropin.js
+++ b/src/dropin.js
@@ -449,7 +449,7 @@ Dropin.prototype.updateConfiguration = function (property, key, value) {
  * * `venmo`
  *
  * @public
- * @returns {string[]} An array of possible payment optoins.
+ * @returns {string[]} An array of possible payment options.
  * @example
  * var paymentOptions = dropinInstance.getAvailablePaymentOptions(); // ['card', 'venmo', 'paypal']
  *
--- a/src/dropin.js
+++ b/src/dropin.js
@@ -440,6 +440,28 @@ Dropin.prototype.updateConfiguration = function (property, key, value) {
 };
 
 /**
+ * Get a list of the available payment methods presented to the user. This is useful for knowing if a paricular payment option was presented to a customer that is browser dependant such as Apple Pay, Google Pay and Venmo. Returns an array of strings. Possible values:
+ * * `applePay`
+ * * `card`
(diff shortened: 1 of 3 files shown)


src/dropin.js
```diff
@@ -439,6 +439,28 @@ Dropin.prototype.updateConfiguration = function (property, key, value) {
   this._navigateToInitialView();
 };
 
+/**
+ * Get a list of the available payment methods presented to the user. This is useful for knowing if a paricular payment option was presented to a customer that is browser dependant such as Apple Pay, Google Pay and Venmo. Returns an array of strings. Possible values:
```
scannillo, line 443: Is there a limitation preventing them from calling `getAvailablePaymentOptions()` before the drop-in UI appears?

src/dropin.js
```diff
@@ -439,6 +439,28 @@ Dropin.prototype.updateConfiguration = function (property, key, value) {
...
+ * * `applePay`
+ * * `card`
+ * * `googlePay`
+ * * `paypalCredit`
+ * * `paypal`
+ * * `venmo`
+ *
+ * @public
+ * @returns {string[]} An array of possible payment optoins.
```
scannillo, line 452:
```suggestion
 * @returns {string[]} An array of possible payment options.
```

src/dropin.js
crookedneighbor: Shouldn't be a problem, since you can't access this method until the instance is ready.

PR #639, opened 2020-07-23
Finnish Translation Support - Enhancement

AlexanderCollins:
Finnish Translation Support - Enhancement

### Summary

Finnish Translations as per (Paypal supported language locales)[https://developer.paypal.com/docs/api/reference/locale-codes/#supported-locale-codes], translations made by two native Finnish speakers, one with C2 level english certification. 

Requires escalation to Localization team as mentioned in [issue report 641](https://github.com/braintree/braintree-web-drop-in/issues/641).

### Checklist

- [ ] Added a changelog entry

I haven't yet added a change log entry - I thought I'd leave this to the localization team.

- [AlexanderCollins](https://github.com/AlexanderCollins)
- [Oageoni](https://github.com/Oageoni) (translations)
- [Mjturt](https://github.com/mjturt) (translations)

PR #579, opened 2019-11-29
Create fi_FI

futureimage:
We are using braintree also in Finnish on our website www.maskmill.com. Pls apply this Finnish translation to your library.

Just notice the below summary??? If this really applies and you are not able to use the below translation, would be nice if you could show the summary already on the translations page description AND disable new file creation possibility in order to save useless extra work from contributors, thanks!

### Checklist

- [ ] Added a changelog entry

futureimage

PR #444, opened 2018-10-13
Create hu_HU.js

alephlg:
Hungarian translation

### Summary
I want use the Braintree in Hungary and I need a hungarian translated frontend

### Checklist

- First version

PR #422, opened 2018-08-20
Greek translation

pmatsinopoulos:
### Summary

Added translation for Greek.

### Checklist

- [X] Added a changelog entry

PR #668, opened 2020-09-29
Add new locales

### LINKED ISSUE #641: Finnish Translations Support
Hi there!

My Finnish team were in need of some Finnish translations so we made some and thought we would share them here!

I read the contributing guideline and it looks like translation requests are not allowed and also would only be included if the language is listed within [PayPal's accepted language codes](https://developer.paypal.com/docs/api/reference/locale-codes/#supported-locale-codes) which luckily appears to be the case. (fi_FI priority 0)

Regarding the translations themselves, I have a native Finnish speaker with C2 english level proficiency who has completed the translations from english and have also had the translations verified by another native Finnish speaker. I've prepared a [pull request: 639](https://github.com/braintree/braintree-web-drop-in/pull/639).

It would be great if this could be escalated to the localisation team for verification. In the mean time we will use the translations directly with the translations prop as per the [braintree-web-drop-in documentation](https://braintree.github.io/braintree-web-drop-in/docs/current/module-braintree-web-drop-in.html#create-examples).

Many Thanks!

crookedneighbor:
closes #187
closes #275
closes #300
closes #334
closes #401
closes #422
closes #444
closes #450
closes #560
closes #579
closes #639
closes #641

### Summary

Adds missing translations.

### Checklist

- [x] Added a changelog entry

- @crookedneighbor


src/translations/index.js
```diff
@@ -22,28 +31,36 @@ var fiveCharacterLocales = {
...
 var twoCharacterLocaleAliases = {
+  ar: fiveCharacterLocales.ar_EG,
+  cs: fiveCharacterLocales.cs_CZ,
   da: fiveCharacterLocales.da_DK,
   de: fiveCharacterLocales.de_DE,
+  el: fiveCharacterLocales.el_GR,
   en: fiveCharacterLocales.en_US,
   es: fiveCharacterLocales.es_ES,
+  fi: fiveCharacterLocales.fi_FI,
   fr: fiveCharacterLocales.fr_FR,
```
scannillo, line 52: Curious - how do you choose which localization gets the 2 letter alias. For example for `fr`, how did you decide to use `fr_FR` instead of `fr_XC`. Are these standards listed somewhere for reference?

crookedneighbor: We prefer the base language over the variant language when a merchant selects the locale using only 2 characters. So if `fr` is passed as the locale, we assume they want `fr_FR` rather than the variant.

scannillo: Thanks!

scannillo (approved):

sshropshire (approved):

PR #675, opened 2020-11-16
Add `showDefaultPaymentMethodFirst` option to create

### LINKED ISSUE #652: `options.preselectVaultedPaymentMethod` does not preselect the active vaulted payment method
### General information

* SDK/Library version: 1.23.0
* Environment: sandbox and production
* Browser and OS: any

### Issue description

When initializing with `braintree.dropin.create` with a token connected to a customer ID the [`preselectVaultedPaymentMethod` is `true` by default](https://braintree.github.io/braintree-web-drop-in/docs/current/module-braintree-web-drop-in.html#.create), but a different card shows at the top of the UI and is selected than the users active payment method.

The below is what I would expect and to me this seems like an issue... or I'm doing something wrong. Would someone clarify if this is expected?

**Expected**:
The users active payment method should show selected at the top of the UI list of payment methods.

**Actual**:
A random payment method is selected and shown at the top of the list of payment methods in the UI.

> payload from `gateway.paymentMethod.find(token)`
> https://developers.braintreepayments.com/reference/request/payment-method/find/node
Screen Shot 2020-08-22 at 2 13 25 PM

> issue (note the text we add from the user's active payment method conflicting with the selected card.
Screen Shot 2020-08-22 at 2 13 51 PM

crookedneighbor:
closes #652

### Summary

Allows whether to show the default payment method first when displaying vaulted payment methods to be configured.

### Checklist

- [x] Added a changelog entry

- @crookedneighbor

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ CHANGELOG
 unreleased
 ----------
 - Fix issue where payment method text may be clipped
-- Add ability to let the Braintree API decide which payment method to show first (typically, the last payment method added to the customer) instead of showing the default payment method first
+- Add ability to let the Braintree API decide which payment method to show first (typically, the last payment method added to the customer) instead of showing the payment method designated as the customer's default in the Braintree control panel
   ```js
   braintree.dropin.create({
     // other config options
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ CHANGELOG
 unreleased
 ----------
 - Fix issue where payment method text may be clipped
+- Add ability to let the Braintree API decide which payment method to show first (typically, the last payment method added to the customer) instead of showing the default payment method first
+  ```js
+  braintree.dropin.create({
+    // other config options
+    showDefaultPaymentMethodFirst: false
+  });
+  ```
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ CHANGELOG
 unreleased
 ----------
 - Fix issue where payment method text may be clipped
+- Add ability to let the Braintree API decide which payment method to show first (typically, the last payment method added to the customer) instead of showing the default payment method first
```
scannillo, line 7: I think this wording is a bit confusing. What determines a `default payment method`?

crookedneighbor: Good callout. So the customer's default payment method is the _first_ payment method added to the customer, or whatever payment method is made the default using the `Custromer: update` methods or set as default in the Braintree control panel.

How about this:

```suggestion
- Add ability to let the Braintree API decide which payment method to show first (typically, the last payment method added to the customer) instead of showing the payment method designated as the customer's default in the Braintree control panel
```

scannillo: This update is helpful! Thanks for clarifying

billwerges (approved):

PR #680, opened 2020-12-21
Fixup apple pay usage of canMakePaymentsWithActiveCard

crookedneighbor:
### Summary

The requirements for the `canMakePaymentsWithActiveCard` require Apple Pay to be the only method positioned if it resolves with true. See https://developer.apple.com/documentation/apple_pay_on_the_web/applepaysession/1778000-canmakepaymentswithactivecard

### Checklist

- [x] Added a changelog entry

- @crookedneighbor

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ unreleased
     showDefaultPaymentMethodFirst: false
   });
   ```
+- Apple Pay
+  - Fix issue where Apple Pay view was using `canMakePaymentsWithActiveCard` incorrectly
 
 1.25.0
 ------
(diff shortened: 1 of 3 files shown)


scannillo:
Should we add a test for if the merchant has say Cards, PayPal, & ApplePay enabled and `canMakePaymentsWithActiveCard == true` to verify that the Card & PayPal options don't appear?

Also - do we _have_ to abide by this guideline just yet 🙃 ? Is Apple rejecting apps for this yet?

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ unreleased
     showDefaultPaymentMethodFirst: false
   });
   ```
+- Apple Pay
+  - Fix issue where Apple Pay view was using `canMakePaymentsWithActiveCard` incorrectly
```
scannillo, line 15: Take it our leave it comment - is it worth noting that we're removing the use of `canMakePaymentsWithActiveCard` entirely?

sestevens (approved):

CHANGELOG.md
crookedneighbor: I think it's fine to keep it vague. Most merchants using Drop-in are using it so they don't actually have to think about how to set up things like Apple Pay. If you feel strongly about it, I can certainly change it.

scannillo (approved):

PR #704, opened 2021-03-03
Fix race condition

crookedneighbor:
### Summary

The way Drop-in was constructed, it was possible to get into a race condition during set up where some dependencies (payment options, data collector, 3ds, etc) can report that they are ready before others even start setting up, causing Drop-in to attempt to finish the setup too early and fail (or miss some payment options).

This PR lets Drop-in know up front what dependencies it is setting up and periodically checks if all the dependencies have reached a terminal set up state (either done, failed or not enabled).

### Checklist

- [x] Added a changelog entry

- @crookedneighbor 
- @jplukarski

### DIFF
--- a/src/dropin-model.js
+++ b/src/dropin-model.js
@@ -6,6 +6,7 @@ var EventEmitter = require('@braintree/event-emitter');
 var constants = require('./constants');
 var paymentMethodTypes = constants.paymentMethodTypes;
 var paymentOptionIDs = constants.paymentOptionIDs;
+var dependencySetupStates = constants.dependencySetupStates;
 var isGuestCheckout = require('./lib/is-guest-checkout');
 var Promise = require('./lib/promise');
 var paymentSheetViews = require('./views/payment-sheet-views');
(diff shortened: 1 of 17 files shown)


sshropshire (approved):

src/dropin-model.js
```diff
@@ -219,29 +249,32 @@ DropinModel.prototype.reportAppSwitchError = function (sheetId, error) {
   };
 };
 
-DropinModel.prototype.asyncDependencyStarting = function () {
-  this.dependenciesInitializing++;
+DropinModel.prototype.hasAtLeastOneAvailablePaymentOption = function () {
+  var self = this;
+  var i;
+
+  console.log(this.supportedPaymentOptions);
```
sshropshire, line 256: These console logs look like they can be removed.

src/dropin-model.js
```diff
@@ -50,13 +63,30 @@ EventEmitter.createChild(DropinModel);
 
 DropinModel.prototype.initialize = function () {
   var self = this;
+  var dependencyReadyInterval = setInterval(function () {
+    var ready = true;
+
+    ASYNC_DEPENDENCIES.forEach(function (dep) {
```
billwerges, line 69: Do we need to iterate through all possible `ASYNC_DEPENDENCIES` here, or can we only iterate the filtered `self.dependencyStates` array?

billwerges (approved):

src/dropin-model.js
crookedneighbor: Yep, nice catch.

src/dropin-model.js
crookedneighbor: You're totally right. Let's do that. 

```suggestion
    Object.keys(self.dependencyStates).forEach(function (dep) {
```

src/dropin-model.js
```diff
@@ -50,13 +63,30 @@ EventEmitter.createChild(DropinModel);
...
+
+    Object.keys(self.dependencyStates).forEach(function (dep) {
+      if (self.dependencyStates[dep] === dependencySetupStates.INITIALIZING) {
+        ready = false;
+      }
+    });
+
+    if (!ready) {
+      return;
+    }
```
billwerges, line 77: We may be able to condense this whole block down to:

```
if(Object.values(self.dependencyStates).includes(dependencySetupStates.INITIALIZING)) {
    return;
}
```

crookedneighbor: Yeah, that looks much better, but we can't do it currently because IE 11 does not support `Object.values`.

Definitely something to look into for the next major version when we'll be transpiling from Typescript anyway.

vaultwiki: Micro-optimization: If you use a for-var-in-loop here rather than the forEach, you can return early as soon as INITIALIZING is seen, rather than looping through every dependency every time or even using a ready-var. As far as condensed-ness goes, it's more condensed than the current code.

PR #709, opened 2021-03-13
Previously discussed micro-optimization in race-condition fix

crackedeggs1:
Finally got around to rebasing my fork, so I can finally pull request the micro-optimization as discussed over here: https://github.com/braintree/braintree-web-drop-in/pull/704#discussion_r588787000

The change eliminates at least 2 function/method calls, reduces the number of times a loop is iterated, and is more compact than the original.


crookedneighbor:
Thanks for doing this. Running the integration tests locally now.

crookedneighbor:
Thanks! Merged in via the command line.

PR #722, opened 2021-04-19
Accessibility updates

jplukarski:
### Summary

Improves keyboard navigation

### Checklist

- [X] Added a changelog entry

- @crookedneighbor 
- @jplukarski


src/views/payment-methods-view.js
```diff
@@ -106,7 +106,7 @@ PaymentMethodsView.prototype._addPaymentMethod = function (paymentMethod) {
   });
 
   if (this.model.isGuestCheckout && this.container.firstChild) {
-    this.container.removeChild(this.container.firstChild);
+    this.views[0].teardown();
     this.views.pop();
```
sshropshire, line 110: I know this may be out of scope, but shouldn't we be using`shift()` instead of `pop()` here?

sshropshire (approved):

src/views/payment-methods-view.js
jplukarski: Maybe @crookedneighbor can offer further clarification, but my guess is at this point we always expect the length of `this.views` to be 1 since this is the guest checkout flow and `pop()` offers some performance enhancements over `shift()` since `shift()` must shift the remaining elements to the left once the first element is removed. Although in this case, since there is only 1 element in the array I am not 100% certain if there are any improvements.

sshropshire: @jplukarski got it yeah if the array has only one element then shift and pop have the same function. That makes sense to me.

crookedneighbor: Yeah, in this particular case, it's identical.

Issue #724, opened 2021-04-23
Vertical scrollbar appearing for saved payment method


dereksdev:
### General information

* SDK/Library version: 1.27.0
* Environment:  Sandbox (maybe Production?)
* Browser and OS: Chrome 90.0.4430.85/Windows 10

### Issue description

A vertical scrollbar appears in saved payment method next to the checkmark on the right side of the label. Removing the `overflow-x: hidden;` CSS rule resolves the issue, but not sure why it is there in the first place.

![scrollbars](https://user-images.githubusercontent.com/4105471/115933271-650c4880-a443-11eb-82a4-bf5d83cea341.png)

crookedneighbor:
Thanks for the report, we'll look into it.

crookedneighbor:
This is out in v1.28.0

PR #727, opened 2021-04-27
Fix issue where vertical scrollbar appears for saved payment method

jplukarski:
### Summary

Fix issue where vertical scrollbar appears on saved payment methods ( #724 )

### Checklist

- [x] Added a changelog entry

### Authors
- @crookedneighbor 
- @jplukarski

### DIFF
--- a/src/less/main.less
+++ b/src/less/main.less
@@ -758,6 +758,7 @@
     text-align: left;
     text-overflow: ellipsis;
     overflow-x: hidden;
+    overflow-y: hidden;
 
     .braintree-method__label--small {
       font-size: 13px;
--- a/src/less/main.less
+++ b/src/less/main.less
@@ -757,8 +757,7 @@
     margin-left: 20px;
     text-align: left;
     text-overflow: ellipsis;
-    overflow-x: hidden;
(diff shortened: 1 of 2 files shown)


sshropshire (approved):
🌊

src/less/main.less
```diff
@@ -758,6 +758,7 @@
     text-align: left;
     text-overflow: ellipsis;
     overflow-x: hidden;
+    overflow-y: hidden;
```
crookedneighbor, line 761:
```suggestion
    overflow: hidden;
```

crookedneighbor:
Just leaving a note here that we had this code this way for a while before making a change to have it be just `overflow-x` https://github.com/braintree/braintree-web-drop-in/commit/8f0cecf750551a99e438a717c30a93de6ab9ffc9

This was because a merchant had some styles on their page that caused some of the text to get clipped off at the y access as a result. Obviously, showing this scrollbar is much worse than a tiny bit of text getting cut off, so we're reverting back to `overflow: hidden`, but wanted to note that in this pr so it's easy to track down later if it comes up again.

PR #756, opened 2021-07-30
fix: add oncancel handler for Apple Pay payment sheet

crookedneighbor:
### Summary

Currently, if a customer cancels the Apple Pay flow, it puts the button in an unreachable state. This fix handles that scenario.

### Checklist

- [x] Added a changelog entry

- @crookedneighbor

### DIFF
--- a/src/views/payment-sheet-views/apple-pay-view.js
+++ b/src/views/payment-sheet-views/apple-pay-view.js
@@ -84,6 +84,10 @@ ApplePayView.prototype._showPaymentSheet = function () {
     });
   };
 
+  session.oncancel = function () {
+    self._sessionInProgress = false;
+  };
+
   session.begin();
 
   return false;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,12 @@
 CHANGELOG
 =========
 
-1.31.0
+unreleased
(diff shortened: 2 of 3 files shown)


sshropshire (approved):

src/views/payment-sheet-views/apple-pay-view.js
```diff
@@ -84,6 +84,10 @@ ApplePayView.prototype._showPaymentSheet = function () {
     });
   };
 
+  session.oncancel = function () {
```
sshropshire, line 87: Interesting choice by Apple to go all lower case for event handlers.

crookedneighbor: Yeah, it looks weird, doesn't it? But it's the pattern the DOM APIs have had from the beginning: https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onclick

We just almost never see it because most people do:

```js
input.addEventListener('click', () => {});
```

instead of 

```js
input.onclick = () => {};
```

Honestly, if Apple provided an `addEventListener` alternative, I'd do it just to not have to deal with how ugly `oncancel` looks :)

jplukarski (approved):

PR #759, opened 2021-08-10
fix: correct issue where Drop-in could get into unusable state

crookedneighbor:
### Summary

Fixes race condition where it was possible to go to the methods view when no payment methods were actually available.

Discovered because a merchant attempted to call `clearSelectedPaymentMethod` in the callback after calling `requestPaymentMethod` on the card view.

Basically, the card view involves some extra animations compared to the other payment methods. So it was possible to put Drop-in into a state where `requestPaymentMethod` had resolved, but the UI had not caught up the action. When something happens to clear out the saved payment methods (such as `clearSelectedPaymentMethod`), then the async action to finish animating the card form -> methods view transition occurs and attempts to put Drop-in into the methods view.... which no longer has any payment methods to display.

We tried a few different approaches for how to handle this, and settled on this one to go to a different view when the methods view was not actually available.

### Checklist

- [x] Added a changelog entry

- @crookedneighbor 
- @jplukarski

### DIFF
--- a/test/unit/dropin.js
+++ b/test/unit/dropin.js
@@ -2106,7 +2106,7 @@ describe('Dropin', () => {
     );
 
     test(
-      'sets primary view to the intial view if on the methods view and there are not saved payment methods and only one payment option is available',
+      'sets primary view to the initial view if on the methods view and there are no saved payment methods and only one payment option is available',
       () => {
         const instance = new Dropin(testContext.dropinOptions);
         const getViewStub = jest.fn();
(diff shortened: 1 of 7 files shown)


test/unit/dropin.js
```diff
@@ -2115,48 +2106,7 @@ describe('Dropin', () => {
...
-        instance.clearSelectedPaymentMethod();
-
-        expect(instance._mainView.setPrimaryView).toBeCalledTimes(1);
-        expect(instance._mainView.setPrimaryView).toBeCalledWith('options');
-      }
-    );
-
-    test(
-      'sets primary view to available payment option view if on the methods view and there are not saved payment methods and only one payment option is available',
+      'sets primary view to the intial view if on the methods view and there are not saved payment methods and only one payment option is available',
```
sshropshire, line 2109:
```suggestion
      'sets primary view to the intial view if on the methods view and there are no saved payment methods and only one payment option is available',
```

sshropshire (approved):

PR #776, opened 2021-11-02
fix: stop passing card true to CardView

### LINKED ISSUE #775: Passing Card as True into create throws a TypeError
### General information

* SDK/Library version: 1.32.0
* Environment: N/a
* Browser and OS Google Chrome	95.0.4638.54

### Issue description

Pasing `card: true` in the create function like below
```
braintree.dropin.create({
        authorization: client_token,
        card: true
)}
```
Returns the following error
```
TypeError: Cannot create property 'vault' on boolean 'true'
    at CardView.initialize (dropin.js:22051)
    at MainView.<anonymous> (dropin.js:20958)
    at Array.reduce (<anonymous>)
    at MainView._initialize (dropin.js:20945)
    at new MainView (dropin.js:20913)
    at Dropin._setUpDependenciesAndViews (dropin.js:16623)
    at dropin.js:16466
```

The documentation mentions that a Boolean value should be passable. 

In addition in Version 1.17.0, passing true to this function did not return this error but in version 1.18.0 this started erroring. 

It is helpful to pass a boolean value here as having to conditionally pass a key value adds additional complexity

the133448:
### Summary

Fixes #775

If the merchant explicitly pass through card as true in merchant configuration options then we can ignore the property as card is turned on by default.

Attempting to fix this in the card view is more challenging as the `true` value will cause TypeErrors

### Checklist

- [ ✔] Added a changelog entry

@the133448

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 CHANGELOG
 =========
+1.32.1
+----------
+- Fix issue where passing `true` to card as Merchant Configuration stops creation
 
 1.32.0
 ----------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 CHANGELOG
 =========
-1.32.1
+UNRELEASED
 ----------
 - Fix issue where passing `true` to card as Merchant Configuration stops creation
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "braintree-web-drop-in",
-  "version": "1.32.0",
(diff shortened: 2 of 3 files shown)


crookedneighbor:
Thanks for catching this. I moved the logic to the card view: https://github.com/braintree/braintree-web-drop-in/commit/d80045a88d5a5e608368c5ec77bbd51da4e1c25a

the133448:
Thanks @crookedneighbor, when will you roll out your next release?

crookedneighbor:
We're tentatively looking at next week for a release.

PR #782, opened 2021-12-07
Update CSP documentation for 3DS 2

kmaurisak:
### Summary

Updated jsdoc to include an explanation of CSP implications with 3DS 2.0

### Checklist

- [x] Updated Home.md


jsdoc/Home.md
```diff
@@ -167,16 +167,24 @@ If using [Google Pay](module-braintree-web-drop-in.html#~googlePayCreateOptions)
...
 
-### 3D Secure Specific Directives
+### 3D Secure 1 Specific Directives
 
-If using [3D Secure](module-braintree-web-drop-in.html#~threeDSecureOptions), include these additional directives:
+If using the [3D Secure component](module-braintree-web_three-d-secure.html) with a 3D Secure 1.0 integration, include these additional directives:
 
 |             | Sandbox                           | Production                    |
 |-------------|-----------------------------------|-------------------------------|
 | script-src  | songbirdstag.cardinalcommerce.com | songbird.cardinalcommerce.com |
```
crookedneighbor, line 176: IMO, we shouldn't differentiate between v1 and v2 in our docs here, especially since v2 is now the default for merchants on the backend and we [default to v2 in Drop-in](https://github.com/braintree/braintree-web-drop-in/blob/250061ecb680263266279565fa4b6fa30b451215/src/lib/three-d-secure.js#L23). We should just update the `frame-src` for production to be `*` and have the note about why it is this way below (as you have it)

crookedneighbor (changes requested):
Just a slight change here, then we're good to go!

sshropshire (approved):

jsdoc/Home.md
```diff
@@ -169,14 +169,20 @@ The `style-src` directive is required so that the styles for the Google Pay butt
...
 ### 3D Secure Specific Directives
 
-If using [3D Secure](module-braintree-web-drop-in.html#~threeDSecureOptions), include these additional directives:
+If using the [3D Secure component](module-braintree-web_three-d-secure.html) with a 3D Secure 1.0 integration, include these additional directives:
 
 |             | Sandbox                           | Production                    |
 |-------------|-----------------------------------|-------------------------------|
 | script-src  | songbirdstag.cardinalcommerce.com | songbird.cardinalcommerce.com |
-| frame-src   | &#42;.cardinalcommerce.com        | &#42;.cardinalcommerce.com    |
+| frame-src   | &#42;.cardinalcommerce.com        | frame-src *                   |
```
sshropshire, line 177: random q: does this mean sandbox's CSP is more restrictive than prod?

crookedneighbor: Technically yes, because in sandbox, everything goes through `*.cardinalcommerce.com`'s sandbox. Should probably just set it to be the same for both, just for simplicity. Should also use the html encoded version of `*`

```suggestion
| frame-src   | &#42;                             | &#42;                         |
```

jplukarski: Good catch, the same directive should be specified for sandbox. I made that update.

jsdoc/Home.md
crookedneighbor: Still need these changes to be made, since you can't really _not_ have a 3ds 2 integration in Drop-in.

jsdoc/Home.md
```diff
@@ -174,9 +174,15 @@ If using [3D Secure](module-braintree-web-drop-in.html#~threeDSecureOptions), in
...
+| frame-src   | &#42;                             | &#42;                         |
 | connect-src | &#42;.cardinalcommerce.com        | &#42;.cardinalcommerce.com    |
 
+3D Secure 2 utilizes an iframe implementation that requires the use of the issuing bank's full ACS URL. In contrast to 3D Secure 1, the 3D Secure 2 core framework does not allow masked URLs or redirects. Given that the list of possible ACS URLs changes regularly and varies between issuers and ACS providers, there is not a strict CSP configuration available for 3D Secure 2.
+
+Additionally, 3D Secure 2 includes a data collection flow called "3DS Method" or "Method URL Collection", which also utilizes the ACS URL directly. This process increases authentication success significantly and is considered mandatory by Visa. Blocking this process through a CSP can potentially result in authentication failures and increased friction within the checkout experience.
+
+If maintaining a CSP in an integration that uses 3D Secure 2, merchants can consider setting `frame-src *` to whitelist all potential ACS URLs that could be utilized during the 3D Secure authentication process.
```
crookedneighbor, line 184:
```suggestion
If maintaining a CSP in an integration that uses 3D Secure, merchants must set `frame-src *` to allowlist all potential ACS URLs that could be utilized during the 3D Secure authentication process.
```

crookedneighbor (changes requested):

crookedneighbor (approved):

PR #806, opened 2022-02-24
Fix: Drop-In inputs escape Drop-In container when there are errors on multiple inputs

jplukarski:
### Summary

When multiple validation errors are displayed, the input fields of Drop-In escape the container. This changes the max-height of the Drop-In container to `fit-content` instead of being hardcoded to 500px.

### Checklist

- [x] Added a changelog entry

- @saralvasquez 
- @jplukarski


src/less/main.less
```diff
@@ -377,7 +377,9 @@
   border-radius: @radius-default;
   display: none;
   margin: 0 auto;
-  max-height: 500px;
+  // NEXT_MAJOR_VERSION drop this when IE support is no longer required
+  max-height: 600px; // IE11 doesn't support fit-content, so we need a backup value here
+  max-height: fit-content;
```
jplukarski, line 382: We tried to find context around why this was originally set to 500px but couldn't find it; it seemed almost arbitrary. We would like to know if there are some ramifications to changing this to `fit-content` that we might be missing.

jplukarski:
retest this please

src/less/main.less
crookedneighbor: This note _feels_ right.

PR #812, opened 2022-04-06
Fix hiper icon visibility that spoils card icons

tamtamchik:
### Summary

Since https://github.com/braintree/braintree-web-drop-in/pull/788 list of cards is spoiled because the hiper icon is not really hidden. Just check images before and after this fix:

Before fix:

After fix:
1 33 0-fixed

### Checklist

- [ ] Added a changelog entry


crookedneighbor (approved):

crookedneighbor:
Awesome, thanks for catching this. Hoping to get a release out this week.

PR #839, opened 2022-10-28
chore: remove hint

crookedneighbor:
### Summary

Hint has not worked for a long time in this SDK, and now that we're 2 major versions behind, is the source of a few Dependabot PRs.

This PR removes `hint` as a dependency. (alternatively, we could update the packages, but it would still be fundamentally broken and not really useful as a tool without further work)

@gesa I know you were a proponent of using Hint originally, do you think it's worth it to put in the time to continue using it?

### Checklist

- [ ] ~Added a changelog entry~

- @crookedneighbor

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,3 @@ coverage/
 .DS_Store
 
 .env
-sauce_connect.ready
-
-hint-report/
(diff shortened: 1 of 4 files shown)


gesa (approved):
I think it is worth devoting real time and effort into. Perhaps at the same time we engage a designer and are ready to release a new major version.

In the meantime, I think it's a good idea to remove it altogether.

PR #865, opened 2023-02-10
Load third-party scripts with CORS support

psyklopz:
Load data-collector with CORS support for crash reporting
Load PayPal checkout with CORS support for crash reporting
Load Google Pay with CORS support for crash reporting

### Summary

We've been chasing an issue where Google's pay.js throws an error for certain users.  However, our crash reporting software does not have access to the stack trace as it's blocked by the CORS policy.

Rather than having to rebuild and host our own version of `braintree-web-drop-in` with every update we'd love it if this could be included by default.  I think it adds value for other users who are in a similar situation and monitor their website's performance with a crash reporting tool.

The Braintree asset-loader project already supports setting the `crossorigin='anonymous'` attribute on the inserted script tag, this PR just updates the `braintree-web-drop-in` to pass the appropriate parameter to the call.

I'm open to feedback and willing to adjust this pull request if desired.

### Checklist

- [X] Added a changelog entry

### Authors
- @psyklopz


ibooker:
We're currently looking into this issue.  I

ibooker:
@psyklopz Have you successfully tested with these changes? Based on our testing, it appears that the `pay.google.com` domain does not support CORS.

nidhinarendra (changes requested):
We need a specific way to test these changes. Please include the testing steps and unit tests for the same.

nidhinarendra:
Since we have not got any response, we are closing this request. If you have any questions later please reach out to us.

PR #870, opened 2023-03-14
chore: remove dependency on class-list module

crookedneighbor:
### Summary

No longer need this module because all the browsers we support have access to the native api.

### Checklist

- [x] Added a changelog entry

- @crookedneighbor

### DIFF
--- a/src/views/payment-sheet-views/card-view.js
+++ b/src/views/payment-sheet-views/card-view.js
@@ -3,7 +3,6 @@
 var assign = require('../../lib/assign').assign;
 var fs = require('fs');
 var BaseView = require('../base-view');
-var classList = require('@braintree/class-list');
 var constants = require('../../constants');
 var DropinError = require('../../lib/dropin-error');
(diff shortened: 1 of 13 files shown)


src/views/payment-sheet-views/card-view.js
```diff
@@ -73,7 +72,7 @@ CardView.prototype.initialize = function () {
   }
 
   if (!this.model.isGuestCheckout && this.merchantConfiguration.vault.allowVaultCardOverride === true) {
-    classList.remove(this.getElementById('save-card-field-group'), 'braintree-hidden');
+    this.getElementById('save-card-field-group').classList.remove('braintree-hidden');
```
bywood, line 75: Is this `getElementById` guaranteed to not return `null`?

bywood: Checked `@braintree/class-list` and it looks like it doesn't do a check there, so this is probably okay.

bywood (approved):

cgdibble (approved):

PR #878, opened 2023-03-31
docs: fix jsdoc version links and semantics

cgdibble:
### Summary

We were getting broken links in the built docs because we have a placeholder for the SDK version that gets swapped out for the _actual_ version at build time. Some of the JSDoc comments are large enough that ended up with more than one version placeholder, thus a basic use of `replace` wasn't going to work since it[ just does the first occurrence](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace).

Specifically:
> If pattern is a string, only the first occurrence will be replaced. 

So, a basic regex does the trick.

During this exploration, I also noted that the JSDocs were using incorrect semantic JSDoc syntax. For a bunch of type definitions we were using `@param` instead of the appropriate `@property`: https://jsdoc.app/tags-typedef.html

### Checklist

- [ ] Added a changelog entry

-

### DIFF
--- a/jsdoc/version-interpolator-plugin.js
+++ b/jsdoc/version-interpolator-plugin.js
@@ -3,6 +3,6 @@ var BT_WEB_VERSION = require('braintree-web').VERSION;
 
 exports.handlers = {
   jsdocCommentFound: function(e) {
-    e.comment = e.comment.replace('{@pkg version}', VERSION).replace('{@pkg bt-web-version}', BT_WEB_VERSION);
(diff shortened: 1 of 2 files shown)


ibooker (commented):
lgtm

ibooker (approved):

shango420:
cdb906ed4a3b21b948cf62636cd661d9c24db464

scannillo (approved):

PR #885, opened 2023-05-22
Fix: Removing card icons if they are overridden

oscarleonnogales:
### Summary

Currently, if a card type is overridden in the `card.overrides.fields.number.supportedCardBrands` object, the buyer sees an error message during card validation, but the card icon is still being shown. This change adds the `braintree-hidden` class to any card type that's being overridden by the merchant.

Ticket for internal tracking: https://paypal.atlassian.net/browse/LI-10689

### Checklist

- [ ] Added a changelog entry

-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # CHANGELOG
 
 ## Unreleased
-  - No longer showing card icons for overridden card types.
+  - Fix issue where supportedCardBrands overrides were incorrectly showing images for hidden card brands
 
 ## 1.37.0
   - Drop dependency on `promise-polyfill`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## Unreleased
+  - No longer showing card icons for overridden card types.
+
 ## 1.37.0
   - Drop dependency on `promise-polyfill`
   - Drop dependency on @braintree/class-list
--- a/src/views/payment-sheet-views/card-view.js
+++ b/src/views/payment-sheet-views/card-view.js
@@ -29,7 +29,6 @@ CardView.ID = CardView.prototype.ID = constants.paymentOptionIDs.card;
 CardView.prototype.initialize = function () {
   var cvvFieldGroup, postalCodeFieldGroup, hfOptions;
   var cardholderNameGroup = this.getElementById('cardholder-name-field-group');
-  var cardIcons = this.getElementById('card-view-icons');
 
   // If merchant explicty passes a value of `true` for card configuration,
   // we need to treat it as if no card configuration was passed, and provide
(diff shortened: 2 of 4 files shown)


SancSalix:
Don't forget a changelog entry!

src/views/payment-sheet-views/card-view.js
```diff
@@ -114,6 +113,33 @@ CardView.prototype._sendRequestableEvent = function () {
...
 
+CardView.prototype._renderCardIcons = function () {
+  var overrides = this.merchantConfiguration.overrides;
+  var cardIcons = this.getElementById('card-view-icons');
+  var supportCardBrands;
+
+  cardIcons.innerHTML = cardIconHTML;
+
+  try {
+    supportCardBrands = overrides.fields.number.supportedCardBrands;
```
hollabaq86, line 124: should the name of this variable match what we're setting from overrides? Currently it's `supportCardBrands`, I'm wondering if it should be `supportedCardBrands`

src/views/payment-sheet-views/card-view.js
```diff
@@ -114,6 +113,33 @@ CardView.prototype._sendRequestableEvent = function () {
...
+        var selector, iconDiv;
+
+        if (value === false) {
+          selector = 'div[data-braintree-id="' + constants.cardTypeIcons[cardBrand] + '-card-icon"]';
+          iconDiv = document.querySelector(selector);
+          hideCardIcon(iconDiv);
+        }
+      });
+    }
+  } catch (error) {
```
hollabaq86, line 138: do we think it's worth sending an analytic log if we ever hit this error block?

src/views/payment-sheet-views/card-view.js
```diff
@@ -44,7 +43,7 @@ CardView.prototype.initialize = function () {
   this.cardholderNameRequired = this.hasCardholderName && this.merchantConfiguration.cardholderName.required === true;
   hfOptions = this._generateHostedFieldsOptions();
 
-  cardIcons.innerHTML = cardIconHTML;
+  this._renderCardIcons();
```
hollabaq86, line 46: I love that this work is a refactor of some logic into a separate function 😍

hollabaq86 (changes requested):
great first pass! Some questions on analytics and var names. Test coverage LGTM. And ofc, don't forget CHANGELOG 😄

src/views/payment-sheet-views/card-view.js
hollabaq86: Talked through the implementation. Since many merchants don't set overrides, we really should be swallowing this error. And from a readability standpoint, this reads better than checking if there's overrides, _and_ fields, _and_ number, etc etc.

We'll add a comment to add some context on why we're swallowing errors in the try/catch block and should be good to go from there.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## Unreleased
+  - No longer showing card icons for overridden card types.
```
hollabaq86, line 4:
```suggestion
  - Fix issue where supportedCardBrands overrides were incorrectly showing images for hidden card brands
```

hollabaq86: Slight tweak to language to make it a little more clear to folks outside our team

hollabaq86 (approved):
Love these test cases, small tweak to CHANGELOG that you can take or leave.

jplukarski (approved):

PR #887, opened 2023-06-09
Accessibility tags

oscarleonnogales:
### Summary

Added `role="alert"` to the divs containing the temporary error message when a user is filling in a card input field incorrectly.

For internal tracking, 7641

### Checklist

- [X] Added a changelog entry

- @oscarleonnogales

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## Unreleased
+ - Added alert role to error message divs beneath card inputs
+
 ## 1.38.0
   - Fix issue where supportedCardBrands overrides were incorrectly showing images for hidden card brands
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## Unreleased
+ - Added alert role to error message divs beneath card inputs
```
jplukarski, line 4: A couple nitpicky suggestions for the Changelog:
- Looks like we use present tense verbs throughout the changelog. So in this case `Added` -> `Add`
- We usually add a reference to the issue we are addressing

```suggestion
 - Add alert role to error message divs beneath card inputs. (#845 [https://github.com/braintree/braintree-web-drop-in/issues/845])
```

hollabaq86 (approved):
LGTM! Linking [Mozilla's docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role) for interested folks

PR #893, opened 2023-07-07
chore: Update braintree-web to 3.96.0

ibooker:
### Summary

Updated `braintree-web` dependency to 3.96.0

### Checklist

- [x] Added a changelog entry

- @ibooker

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## Unreleased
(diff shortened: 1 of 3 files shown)


jplukarski (approved):

hollabaq86 (changes requested):
there's a rather large difference in package-lock files. Can we double check that's expected?

hollabaq86 (approved):

PR #899, opened 2023-07-25
Ca compliance notice of collection

oscarleonnogales:
### Summary

[Jira Ticket](https://paypal.atlassian.net/browse/DTBTSDK-2864)
Add CA compliance notice of collection.

### Checklist

- [X] Added a changelog entry

- @ibooker
- @oscarleonnogales 

Screenshot 2023-07-26 at 9 31 12

Screenshot 2023-07-26 at 9 31 34


src/html/main.html
```diff
@@ -219,6 +219,10 @@
               </label>
               <div data-braintree-id="postal-code-field-error" class="braintree-form__field-error" role="alert"></div>
             </div>
+
+            <div class="braintree-form__notice-of-collection">
+              <a href="https://www.paypal.com/us/legalhub/home?locale.x=en_US" target="_blank" rel="noopener noreferrer">{{noticeOfCollection}}</a>
```
scannillo, line 224: Do we want to avoid hard coding the local to always be US?

src/less/main.less
```diff
@@ -571,6 +571,12 @@
 .braintree-sheet__content--form {
   padding: 10px 15px 10px 10px;
 
+  .braintree-form__notice-of-collection {
+    font-size: 14px;
```
scannillo, line 575: Just my two cents, but I would make this font size smaller. I was thinking similar to the `Having trouble logging in?` text size relative to the fields in the below example:

netflix

src/html/main.html
oscarleonnogales: We asked Arpan about this and were told we don't need to map the URL to the locale.

scannillo: Got it, thanks for confirming! Might be safer just to hard code `https://www.paypal.com/us/legalhub/home` and let the site handle defaulting based on user location

src/less/main.less
oscarleonnogales: Reduced to 12px, updated the screenshots in the description.

src/html/main.html
oscarleonnogales: Great catch!

scannillo (approved):
LGTM!

Only note is to check that the accessibility label & type of the link reads properly as a clickable element.

PR #910, opened 2023-11-14
Add aria-hidden to generic card icon and aria-required to Hosted Fields

jplukarski:
### Summary

Adds the `aria-hidden` label to the generic card icon in the upper left corner of the cardview so that screenreaders do not read it. Also, adds the `aria-required` label to the input in the hosted fields so screen readers can accurately indicate it is a required field.

### Checklist

- [x] Added a changelog entry

- @jplukarski

### DIFF
--- a/test/unit/views/payment-sheet-views/card-view.js
+++ b/test/unit/views/payment-sheet-views/card-view.js
@@ -726,7 +726,7 @@ describe('CardView', () => {
       });
     });
 
-    test.only('sets aria-required attribute on hosted fields', () => {
+    test('sets aria-required attribute on hosted fields', () => {
       fakeClient.getConfiguration.mockReturnValue({
         gatewayConfiguration: {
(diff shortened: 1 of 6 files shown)


test/unit/views/payment-sheet-views/card-view.js
```diff
@@ -726,6 +726,61 @@ describe('CardView', () => {
       });
     });
 
+    test.only('sets aria-required attribute on hosted fields', () => {
```
ibooker, line 729: Should this still say `.only`?

jplukarski: Good catch. Fixed.

ibooker (approved):
Looks good

oscarleonnogales (approved):

jplukarski:
> Change looks good, but noticed the test isn't passing

Yep, fixed the failing unit and publishing tests!

Issue #805, opened 2022-02-17
paymentMethodRequestable/paymentMethodIsSelected with 3DS


ae-michal-mordarski:
### General information

* SDK/Library version: 1.33.0
* Environment: Sandbox
* Browser and OS: Latest version of Chrome/Windows 10 Pro

### Issue description

Per documentation paymentMethodIsSelected will be true any time a payment method is visibly selected in the Drop-in UI, however with 3DS enabled you get false first as expected, followed by true right away, but 3DS pop-up is shown, so technically speaking payment  method is not visibly selected until 3DS is completed. 

Furthermore, we have another logic that depends on paymentMethodIsSelected being true when saved credit card is selected, thus making it impossible to distinguish, because true also comes while 3DS pop-up is visible.

https://braintree.github.io/braintree-web-drop-in/docs/current/Dropin.html#~paymentMethodRequestablePayload

1. Enter credit cards details which requires 3DS
2. requestPaymentMethod (3DS pop-up is shown and card is not selected)

Expected result: paymentMethodIsSelected is true when 3DS is completed
Actual result: paymentMethodIsSelected is true when 3DS popup is shown

```
instance.on('paymentMethodRequestable', function (event) {
	if (!event.paymentMethodIsSelected) {
		console.log('requestPaymentMethod')
		instance.requestPaymentMethod(
			{
				threeDSecure: getThreeDSecure(),
			},
			function (err, payload) {
				if (err) {
					instance.clearSelectedPaymentMethod();
					return;
				}

				console.log(payload);
			},
		);
	}
	else
	{
		console.log('paymentMethodIsSelected')	
	}
});
```

jplukarski:
Hey @ae-michal-mordarski , thank you for reporting this and your patience while we look into this. We agree that it makes sense for the `paymentMethodRequestable` event to not report `paymentMethodIsSelected` as `true` until the 3DS challenge is completed. We have this task on our radar and will update this issue once the fix is in place.

ae-michal-mordarski:
@jplukarski nice, I'm glad I'm able to help this way.

jplukarski:
This bug has been fixed in version 1.42.0.

PR #917, opened 2023-12-18
Emit paymentMethodRequestable event after 3DS Challenge is completed

jplukarski:
### Summary

[Jira](https://paypal.atlassian.net/browse/LI-29388)

Holds off sending the [`paymentMethodRequestable`](https://braintree.github.io/braintree-web-drop-in/docs/current/Dropin.html#event:paymentMethodRequestable) event until after `verifyCard` has completed for 3DS flows.

Addresses #805 

Testing strategy:

Run the [Drop-in demo in development mode](https://github.com/braintree/braintree-web-drop-in/blob/main/DEVELOPMENT.md#demo-app-and-jsdocs) and add the following query parameters to the URL in your browser:

`?threeDSecure=true&authorization=<generate-a-client-token>`

Then test [each test card listed in our 3DS documentation](https://developer.paypal.com/braintree/docs/guides/3d-secure/testing-go-live/ruby) to make sure everything works as expected on each flow. Open the dev console and notice that the event does not fire off until 3DS authentication is completed:

`paymentMethodRequestable {paymentMethodIsSelected: true, type: 'CreditCard'}`

### Checklist

- [X] Added a changelog entry

- @jplukarski

### DIFF
--- a/src/dropin-model.js
+++ b/src/dropin-model.js
@@ -52,6 +52,7 @@ function DropinModel(options) {
   this.failedDependencies = {};
   this._options = options;
   this._setupComplete = false;
+  this._shouldWaitForVerifyCard = false;
 
   while (this.rootNode.parentNode) {
     this.rootNode = this.rootNode.parentNode;
@@ -96,6 +97,10 @@ DropinModel.prototype.initialize = function () {
 
 DropinModel.prototype.confirmDropinReady = function () {
   this._setupComplete = true;
+};
+
+DropinModel.prototype.verifyCardReady = function () {
+  this._shouldWaitForVerifyCard = !this._shouldWaitForVerifyCard;
 };
 
 DropinModel.prototype.isPaymentMethodRequestable = function () {
@@ -220,6 +225,10 @@ DropinModel.prototype._shouldEmitRequestableEvent = function (options) {
     // fixes issues with lazy loading of imports where event
     // should not be emitted
     // https://github.com/braintree/braintree-web-drop-in/issues/511
+    return false;
+  }
+
+  if (this._shouldWaitForVerifyCard) {
     return false;
   }
 
--- a/src/dropin-model.js
+++ b/src/dropin-model.js
@@ -52,7 +52,7 @@ function DropinModel(options) {
   this.failedDependencies = {};
   this._options = options;
   this._setupComplete = false;
-  this._shouldWaitForVerifyCard = false;
+  this.shouldWaitForVerifyCard = false;
 
   while (this.rootNode.parentNode) {
     this.rootNode = this.rootNode.parentNode;
@@ -97,10 +97,6 @@ DropinModel.prototype.initialize = function () {
 
 DropinModel.prototype.confirmDropinReady = function () {
   this._setupComplete = true;
-};
-
-DropinModel.prototype.verifyCardReady = function () {
-  this._shouldWaitForVerifyCard = !this._shouldWaitForVerifyCard;
 };
 
 DropinModel.prototype.isPaymentMethodRequestable = function () {
@@ -228,7 +224,7 @@ DropinModel.prototype._shouldEmitRequestableEvent = function (options) {
     return false;
   }
 
-  if (this._shouldWaitForVerifyCard) {
+  if (this.shouldWaitForVerifyCard) {
     return false;
   }
 
--- a/src/lib/three-d-secure.js
+++ b/src/lib/three-d-secure.js
@@ -39,8 +39,6 @@ ThreeDSecure.prototype.initialize = function () {
 };
 
 ThreeDSecure.prototype.verify = function (payload, merchantProvidedData) {
-  var self = this;
-
   var verifyOptions = assign({
     amount: this._config.amount
   }, merchantProvidedData, {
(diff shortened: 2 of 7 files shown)


src/dropin-model.js
```diff
@@ -98,6 +99,10 @@ DropinModel.prototype.confirmDropinReady = function () {
   this._setupComplete = true;
 };
 
+DropinModel.prototype.verifyCardReady = function () {
```
jplukarski, line 102: Open to suggestions on the name of this method. Had trouble coming up with a good name. Maybe something like `toggleShouldWaitForVerifyCard` but that seems too complex?

jaxdesmarais: GH is acting wild for me, so posting again. Not sure if this is more Swift-y but typically we would name a method returning a `bool` either something like `canVerifyCard` or `isVerifyCardReady`.

src/dropin-model.js
```diff
@@ -98,6 +99,10 @@ DropinModel.prototype.confirmDropinReady = function () {
...
+  this._shouldWaitForVerifyCard = !this._shouldWaitForVerifyCard;
```
hollabaq86, line 103: maybe I need caffeine, but do I undertand this evaluation correctly?

we are setting the value of `_shouldWAitForVerifyCard` to the opposite of `_shouldWaitForVerifyCard`? 

So false?

hollabaq86: Is this some JS magic that I'm simply forgetting lol

test/unit/dropin.js
```diff
@@ -1212,6 +1212,44 @@ describe('Dropin', () => {
       }
     );
 
+    test('calls verifyCardReady and setPaymentMethodRequestable when 3D secure is called', done => {
```
hollabaq86, line 1215:
```suggestion
    test('calls verifyCardReady and setPaymentMethodRequestable when 3D secure is complete', done => {
```

hollabaq86: let's add an expectation that `setPaymentRequestable` isn't being called before 3ds completes

hollabaq86: I'm thinking we may want tests for the following scenarios specific to 3DS if possible:

lookup and no step up challenge
lookup and subsequent step up challenge

Just to be doubly sure we're confident this event is being sent in the right order

src/lib/three-d-secure.js
```diff
@@ -54,6 +54,7 @@ ThreeDSecure.prototype.verify = function (payload, merchantProvidedData) {
 
   verifyOptions.additionalInformation = verifyOptions.additionalInformation || {};
   verifyOptions.additionalInformation.acsWindowSize = verifyOptions.additionalInformation.acsWindowSize || DEFAULT_ACS_WINDOW_SIZE;
+  this._model.verifyCardReady();
```
hollabaq86, line 57: I don't think this is the correct place to call `verifyCardReady`. This is happening when lookup completes, but that's not taking into account when step up authentication occurs (that happens in the verifyCard function)

hollabaq86: I think this logic is going to be a little more complex to cover both scenarios where 3DS "completed" because it didn't require step up authentication, and where 3DS completes after step up authentication. Happy to talk through this in person, let me know!

hollabaq86 (changes requested):
This is a good first pass, but I don't think we cover all scenarios where 3DS "completes." 

Why does 3DS have to be so complicated :lolsob:

src/dropin-model.js
oscarleonnogales: I think it depends on how we're using and calling the function. If we're calling it inside of an if statement, I think it makes sense to call it `canVerifyCard` or `isVerifyCardReady` since it's returning a boolean.

In this case, it's being called as part of the `_threeDSecure.verify` flow, so I like that original idea of calling it `toggleShouldWaitForVerifyCard` or something similar. `toggleWaitForVerification`?

src/dropin-model.js
oscarleonnogales: It's whatever the opposite value used to be (toggling between true/false)

src/lib/three-d-secure.js
oscarleonnogales: I believe it's being called here to toggle `this._shouldWaitForVerifyCard` back to false. So after it's complete, we should no longer wait for verification. Maybe a clearer way to use this function is to rename it and explicitly pass in a boolean value?

jaxdesmarais: Pardon my mobile brain but is there a reason we need a method at all vs setting the property directly? This may be a JS thing so please excuse my very basic questions!

src/dropin-model.js
jplukarski: Oscar is correct in this was meant to toggle between true/false, however based on discussion we are removing this setter method and instead modifying the `shouldWaitForVerifyCard` property directly

src/lib/three-d-secure.js
jplukarski: @jaxdesmarais this is a great question and after discussion, we have refactored this to set the property directly.

jaxdesmarais (approved):
🚀

hollabaq86 (approved):
chefskiss

PR #940, opened 2024-08-06
Use braintree-web@3.106.0

CJGlitter:
### Summary

Upgrade braintree-web version to latest

### Checklist

- [x] Added a changelog entry

- @CJGlitter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## 1.44.0
+  - Update Braintree web to 3.106.0
+
 ## 1.43.0
   - Update Braintree web dependancies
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## 1.44.0
```
GoogilyBoogily, line 3: This should probably be 1.43.1

CJGlitter: So actually this is bumping from the pre-Fastlane version, so I think it merits a minor version bump. But I'm fine with either one.

ibooker (approved):

PR #964, opened 2025-03-27
Set Override for cross-fetch

jplukarski:
### Summary

Set an override for cross-fetch to ensure it is always on version 4.1.0 throughout the dependency tree

### Checklist

- [ ] Added a changelog entry

- @jplukarski

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,5 +80,8 @@
   "browserslist": [
     "last 2 versions",
     "iOS 8"
-  ]
+  ],
+  "overrides": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -80,5 +80,8 @@
   "browserslist": [
     "last 2 versions",
     "iOS 8"
-  ]
+  ],
+  "overrides": {
```
ibooker, line 84: nice!

ibooker (approved):

